Issue #175, opened 2014-08-11
Deterministic UUIDs


fabiopelosin:
I have been thinking about doing something similar to this for a while and the [xUnique](https://github.com/truebit/xUnique) revived my interest on this feature.

##### Context

The algorithm used by Xcode to generate UUIDs is not known. Xcodeproj currently has a conservative approach at tries to make UUIDs which are similar to those generated by Xcodeproj while checking if a duplicate already exists in the project before using a new one. A duplicate UUID would corrupt the project.

[xUnique](https://github.com/truebit/xUnique) proves that Xcode accepts UUIDs regardless of the length and of the type.

##### Proposal

- Generate longer UUIDs
- Derive the UUID from a hash of the attributes of the new objects and the UUID of the parents. This would greatly reduce the noise when recreating a project from scratch.
- Add a subcommand to “normalize” projects.

##### Alternative

- Prefix UUIDs with a string to namespace the ones generated by Xcodeproj and avoid the need to check for collisions.
- Use a hash of the timestamp for the UUIDs.

mrousavy:
I think I'm experiencing an issue with deterministic UUIDs [here](https://github.com/CocoaPods/CocoaPods/issues/10506), does anyone have an idea what might be wrong here?

PR #185, opened 2014-09-12
[WIP] Add Support for deterministic UUIDs

fabiopelosin:
This is the work in progress patch for https://github.com/CocoaPods/Xcodeproj/issues/175.

The difficult part of this patch are:

- Xcodeproj relies on the UUIDs (`objects_by_uuid`) for its internal reference counting.
- Proxies to avoid retain cycles and to allow to reference objects in another projects consider the UUIDs as strings (instead of using the Xcodeproj reference system) and objects, with this patch, might not have an UUID until saved.

There are alternative approaches like specifying an temporary UUID during the creation of the object and then update it during save according to the final attributes of the object. However, I feel that this approach would be slower and I would prefer to first explore any alternative with the approach proposed in this patch.

```
550 specifications (886 requirements), 3 failures, 0 errors
```

---

Closes https://github.com/CocoaPods/CocoaPods/issues/1118

### DIFF
--- a/lib/xcodeproj/project.rb
+++ b/lib/xcodeproj/project.rb
@@ -55,7 +55,7 @@ module Xcodeproj
     #
     def initialize(path, skip_initialization = false)
       @path = Pathname.new(path).expand_path
-      @objects_by_uuid = {}
+      @objects = []
       @generated_uuids = []
       @available_uuids = []
       unless skip_initialization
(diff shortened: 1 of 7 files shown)


kylef:
Hey, I'm going to close this as nothing is moving in a while. Please re-open when this is ready for merge.

PR #236, opened 2015-01-29
[Constants] Fix COPY_PHASE_STRIP

mrackwitz:
As pointed out in CocoaPods/CocoaPods#3062, the value for the build setting `COPY_PHASE_STRIP` is swapped for debug and release configuration.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Xcodeproj Changelog
 
+## Master
+
+#### Bug Fixes
+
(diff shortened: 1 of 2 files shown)


segiddins:
@mrackwitz aren't these defaults generated programatically now?

mrackwitz:
Nope, that was never merged. At the moment, they are only validated by semi-automated created specs, but those cover only the per-target build settings.

segiddins:
In that case, :+1:

kylef:
@mrackwitz Can you add a CHANGELOG entry.

PR #239, opened 2015-02-10
add .H to HEADER_FILES_EXTENSIONS

banjun:
We have legacy header files named capitalized like `FOO.H`.
Currently `pod install` adds these headers to the Compile Sources build phase.
This PR handles `FOO.H` as a header file and does not add to the Compile Sources, same as manually drag & drop them to Xcode project. (.H files are treated as `C Header` in a project by default)

### DIFF
--- a/lib/xcodeproj/project/object/native_target.rb
+++ b/lib/xcodeproj/project/object/native_target.rb
@@ -412,7 +412,7 @@ module Xcodeproj
 
             extension = File.extname(file.path)
             header_extensions = Constants::HEADER_FILES_EXTENSIONS
-            if header_extensions.include?(extension)
+            if header_extensions.include?(extension.downcase)
               headers_build_phase.files << build_file
             else
               if compiler_flags && !compiler_flags.empty?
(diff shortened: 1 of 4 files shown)


segiddins:
It would probably be better to make extension comparisons case insensitive

banjun:
I think it's more aggressive and reasonable.

lib/xcodeproj/project/object/native_target.rb
```diff
@@ -412,7 +412,7 @@ def add_file_references(file_references, compiler_flags = {})
 
             extension = File.extname(file.path)
```
segiddins: I'd prefer the downcase to be here, but no big deal.

segiddins:
Can we have a CHANGELOG entry and some test coverage?

banjun:
added a test and updated CHANGELOG.

segiddins:
Excellent, thanks @banjun!

banjun:
thank you for review!

PR #244, opened 2015-03-06
Propagate language parameter to ProjectHelper::common_build_settings

mrackwitz:
The Swift build settings never have been applied by CocoaPods, as the primarily used target language couldn't been specified so far. This allows to fix CocoaPods/CocoaPods#3062.


kylef:
Can you add a tests to ensure a correct build setting for Swift is **only** exposed when `:swift` is passed and vice-versa?

mrackwitz:
:+1:

kylef:
Thanks :+1:

Issue #243, opened 2015-03-06
Uninitialized constant exception when using PlistHelper


vincentisambart:
```
$ ruby -rxcodeproj -e "Xcodeproj::PlistHelper"
/Users/vincent-isambart/.rbenv/versions/2.2.0/lib/ruby/gems/2.2.0/gems/xcodeproj-0.22.0/lib/xcodeproj/plist_helper.rb:598:in `<module:DevToolsCore>': uninitialized constant DevToolsCore::Pathname (NameError)
	from /Users/vincent-isambart/.rbenv/versions/2.2.0/lib/ruby/gems/2.2.0/gems/xcodeproj-0.22.0/lib/xcodeproj/plist_helper.rb:542:in `<top (required)>'
	from -e:1:in `<main>'
```

It works fine if you `require "pathname"` before but that should not be needed.

segiddins:
We'd love a PR that add the require to the plist helper!

Fl0p:
reopen please

same error
```
[!] uninitialized constant Xcodeproj::PlistHelper
```
with xcodeproj 1.0.0.beta.3
gems list:
```
flop$ gem list

*** LOCAL GEMS ***

actionmailer (4.2.5, 4.1.0, 4.0.2)
actionpack (4.2.5, 4.1.6, 4.1.0, 4.0.2)
actionview (4.2.5, 4.1.6, 4.1.0)
activejob (4.2.5)
activemodel (4.2.5, 4.1.6, 4.0.2)
activerecord (4.2.5, 4.0.2)
activerecord-deprecated_finders (1.0.4, 1.0.3)
activesupport (4.2.5, 4.1.6, 4.1.0, 4.0.2, 3.2.16)
addressable (2.4.0)
arel (7.0.0, 6.0.3, 4.0.1)
atomic (1.1.99, 1.1.16, 1.1.14)
autoparse (0.3.3)
babelish (0.4.2)
bigdecimal (1.2.7, 1.2.0)
builder (3.2.2, 3.1.4)
bundler (1.11.2, 1.3.5)
bundler-unload (1.0.2)
bunlder (0.0.1)
CFPropertyList (2.3.2, 2.2.8)
claide (1.0.0.beta.2, 0.9.1, 0.7.0, 0.5.0, 0.4.0)
cocoapods (1.0.0.beta.5, 0.39.0, 0.32.1, 0.29.0)
[...]

PR #245, opened 2015-03-09
add missing require

vincentisambart:
PlistHelper was using Pathname but not requiring it.
Issue #243


vincentisambart:
Should I also add an entry in the CHANGELOG.md?
The current CHANGELOG.md is incorrect, with the change for #244 (added after releasing 0.22), written in the changes for 0.22.

kylef:
Thanks for your contribution @vincentisambart. I've sorted the CHANGELOG out along with moving the require to `xcodeproj.rb` so Pathname is loaded across the whole module.

PR #246, opened 2015-03-17
Fix add_system_framework could not find the system framework

logiph:
when I’m trying to create project in command line,  the path is wrong
when add System framework into a target, in my environment (Xcode6.2
Mac OS X 10.10.2) the Constants::LAST_KNOWN_IOS_SDK is 7.1 but in Xcode
Folder， there are only iPhoneOS.sdk and iPhoneOS8.2.sdk, so, I think it's no need to add sdk_version in iPhoneOS


segiddins:
This change doesn't seem right to me...

PR #255, opened 2015-03-29
[Config] 'includes' affect equality.

brianpartridge:
I was running into issues where I would load an xcconfig from disk, edit `includes` and try to save, but the change would not actually be saved.  The issue was the equality check in `save_as`.  Since equality was based on `to_hash`, if there were only changes that didn't affect the hash they would never be saved.
:rainbow:

### DIFF
--- a/lib/xcodeproj/config.rb
+++ b/lib/xcodeproj/config.rb
@@ -61,7 +61,7 @@ module Xcodeproj
     end
 
     def ==(other)
-      other.respond_to?(:to_hash) && other.to_hash == to_hash
+      other.attributes == attributes && other.other_linker_flags == other_linker_flags && other.includes == includes
     end
 
     public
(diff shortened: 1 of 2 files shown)


segiddins:
Why not just add includes to the comparison in ==?

brianpartridge:
I liked that this keeps the equality test simple without multiple `respond_to` checks.  Using `to_s` does more compute work, but it's not significantly more. If we were going to go the route of using `to_hash` and `includes`, I would argue that we should be comparing `attributes` and `other_linker_flags` rather than `to_hash` because of the work that it does. Thoughts?

segiddins:
This seems fine to me. @mrackwitz ?

mrackwitz:
I agree with:
> I would argue that we should be comparing `attributes` and `other_linker_flags` rather than `to_hash`

Basing the implementation of `==` on `to_hash` seemed to be convenient for the tests, but when we call it there explicitly anyway, then I'd pledge for implementing it properly.

brianpartridge:
@mrackwitz done.

mrackwitz:
Looks good! Thanks! :+1:

brianpartridge:
Sweet, thanks.

PR #258, opened 2015-03-31
Fix docs description for set_launch_target

shepting:
There was an extra `target` word in the description.


segiddins:
Thanks @shepting!

PR #269, opened 2015-05-17
[PlistHelper] Show a nice error message when loading fiddle fails

segiddins:


kylef:
Is there a case where fiddle isn't available?

segiddins:
You can compile Ruby without it, and it's not available at all on Rubinius.

kylef:
:ok_hand:

PR #270, opened 2015-05-27
PBXContainerItemProxy proxyType is a numeric field

EliSchleifer:
proxyType is a numeric field not a string. This change corrects the type expected


segiddins:
@EliSchleifer this causes a bunch of spec failures, can you please look into those?

EliSchleifer:
I tried making some additional commits to clean up the type information. I have really zero context in ruby and even less awareness of how to debug these tests. Maybe someone with a little more insight can quickly polish this off. Definitely the current type is wrong. I could just open an Issue but I thought I would try to provide a solution

segiddins:
From trying to rebase this, it seems the plist stuff needs the numeric attributes to be a String anyways, so I'm going to close this for now.

PR #285, opened 2015-07-01
fix incorrect SDKROOT parser

huxingyi:
Hi, I think this function `sdk_version` in native_target.rb can't handle the full SDKROOT path well, for example,
```rb
main_target_settings = {"SDKROOT"=>"/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS8.1.sdk"}

main_target_settings.each{|key, value| target.build_settings("Debug")[key] = value}
main_target_settings.each{|key, value| target.build_settings("Release")[key] = value}

puts target.sdk_version()  # this line will output "." not "8.1"
```
Because of this, the function `add_system_framework` will add the wrong path if I set the SDKROOT.
so, I changed some lines to fix this.

### DIFF
--- a/lib/xcodeproj/constants.rb
+++ b/lib/xcodeproj/constants.rb
@@ -138,13 +138,13 @@ module Xcodeproj
         'MTL_ENABLE_DEBUG_INFO'             => 'NO',
       }.freeze,
       [:ios] => {
-        #'SDKROOT'                           => 'iphoneos',
+        'SDKROOT'                           => `xcode-select -print-path`.chomp('\n') + "/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS" + LAST_KNOWN_IOS_SDK + ".sdk",
       }.freeze,
       [:osx] => {
-        'SDKROOT'                           => 'macosx',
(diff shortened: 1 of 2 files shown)


lib/xcodeproj/constants.rb
```diff
@@ -138,13 +138,13 @@ module Constants
         'MTL_ENABLE_DEBUG_INFO'             => 'NO',
       }.freeze,
       [:ios] => {
-        'SDKROOT'                           => 'iphoneos',
+        'SDKROOT'                           => `xcrun xcode-select --print-path`.strip + "/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS" + LAST_KNOWN_IOS_SDK + ".sdk",
```
kylef: These would be absolute paths and therefore make this fragile and won't allow checking in Pods directory as there will be hard-coded paths specific to Xcode installs.

huxingyi: I get it. Maybe change the SDKROOT is not a good idea. I' ll close this pull request.

PR #291, opened 2015-07-24
#192 xcdatamodeld not handled properly

adamhrz:
- Actually worked correclty if xcdatamodeld is in root of project but not if the data model is inside a folder off the root
- Added code to construct correct path from group.parent.path, group.path and path
- Original code would skip children and current version code since path.exists? for data models in a subdirectory would fail
- [ ] Needs a test to simulate datamodel in a sub-directory to prove this fixes the problem

### DIFF
--- a/lib/xcodeproj/project/object/helpers/file_references_factory.rb
+++ b/lib/xcodeproj/project/object/helpers/file_references_factory.rb
@@ -126,7 +126,11 @@ module Xcodeproj
           # @return [XCVersionGroup] The new reference.
           #
           def new_xcdatamodeld(group, path, source_tree)
-            path = Pathname.new(group.parent.path).realpath().join(group.path).join(path)
+            if group.parent.path
+              path = Pathname.new(group.parent.path).realpath.join(group.path).join(path)
+            else
+              path = Pathname.new(path)
+            end
             ref = group.project.new(XCVersionGroup)
             group.children << ref
             GroupableHelper.set_path_with_source_tree(ref, path, source_tree)
--- a/lib/xcodeproj/project/object/helpers/file_references_factory.rb
+++ b/lib/xcodeproj/project/object/helpers/file_references_factory.rb
@@ -126,7 +126,7 @@ module Xcodeproj
           # @return [XCVersionGroup] The new reference.
           #
           def new_xcdatamodeld(group, path, source_tree)
-            path = Pathname.new(path)
+            path = Pathname.new(group.parent.path).realpath().join(group.path).join(path)
             ref = group.project.new(XCVersionGroup)
             group.children << ref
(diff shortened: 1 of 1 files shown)


lib/xcodeproj/project/object/helpers/file_references_factory.rb
```diff
@@ -126,7 +126,7 @@ def new_file_reference(group, path, source_tree)
           # @return [XCVersionGroup] The new reference.
           #
           def new_xcdatamodeld(group, path, source_tree)
-            path = Pathname.new(path)
+            path = Pathname.new(group.parent.path).realpath().join(group.path).join(path)
```
neonichu: Is there a reason for this not being `group.path.realpath.join(path)`?

neonichu:
Thanks for this! Could you have a look at the test failures and my comment?

kylef:
Thanks for the contribution.

Would you be able to add a test case for this change proving that it fixes the problem?

adamhrz:
Working on it this morning

adamhrz:
I'm having a hard time (due to my lack of Ruby experience and direct usage of xcodeproj) writing a failing test that simulates the issue in the ticket.  I'm using liftoff which was how the issue was exposed and I'm able to repro the issue and prove the change fixes it. Could someone help me with writing the test for xcodeproj? I tried cloning new_xcdatamodeld in file_references_factory_spec.rb but I'm just not clear on how to simulate the datamodel file being copied into a subfolder e.g. Classes/Data

lib/xcodeproj/project/object/helpers/file_references_factory.rb
adamhrz: group.path doesn't have a method realpath since group.path is just a string hence the call to Pathname.new:
```
Pathname.new(group.path).realpath
```
group.parent.path is necessary because group.path returns No such file or directory when we call realpath.

adamhrz: Based on your suggestions though I did update the code as follows:
```
            if group.parent.path
              path = Pathname.new(group.parent.path).realpath.join(group.path).join(path)
            else
              path = Pathname.new(path)
            end
```
With this in place the tests pass again and datamodels in sub-directories still work

adamhrz:
Committed code that passes the old tests and added a TODO item for someone more familiar with xcodeproj to help create a new test to confirm that a datamodel in a sub-directory creates an xcdatamodel configured correctly in the xcodeproj as described above

Eldorado234:
Since https://github.com/CocoaPods/Xcodeproj/pull/427 got merged, I think this PR is not needed anymore. @adamhrz can you check if your problem is fixed on `master`?

dnkoutso:
Closing as this is being inactive for 2 years. Feel free to open a new one providing information on why or why not it should land.

PR #292, opened 2015-07-25
[UUIDGenerator] Reduce the situations in which dupes will be generated

segiddins:
Closes https://github.com/CocoaPods/CocoaPods/issues/3823.
Closes https://github.com/CocoaPods/CocoaPods/issues/3850.


alloy:
Would be good to have a test for these cases that generated clashes to ensure it’s not re-introduced later as part of refactoring.

segiddins:
@alloy agree, I'll try and add them, should I hold off on merging this until then?

PR #298, opened 2015-08-26
Fixed - Parse the available SDKs instead of using the constants

carles:
The current LAST_KNOWN_IOS_SDK is 8.3 but watchOS2 apps require 9.0.
We think it makes more sense to get the latest SDK from the current system rather than blindly assigning from a constant which may not exist in file.

In our case, Xcode 7 beta 6 comes with 9.0 but not with 8.3. So all the system frameworks would be pointing to nowhere.

NOTICE: There's a noticeable impact on performance when parsing the SDKs instead of using the constants. Probably we're doing it in a dumb way and there's a smarter alternative.

### DIFF
--- a/lib/xcodeproj/gem_version.rb
+++ b/lib/xcodeproj/gem_version.rb
@@ -1,5 +1,5 @@
 module Xcodeproj
   # The version of the xcodeproj gem.
   #
-  VERSION = '0.27.0' unless defined? Xcodeproj::VERSION
+  VERSION = '0.27.0-pm' unless defined? Xcodeproj::VERSION
 end
--- a/lib/xcodeproj/constants.rb
+++ b/lib/xcodeproj/constants.rb
@@ -4,11 +4,11 @@ module Xcodeproj
   module Constants
     # @return [String] The last known iOS SDK (stable).
     #
-    LAST_KNOWN_IOS_SDK = '8.3'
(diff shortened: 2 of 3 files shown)


lib/xcodeproj/gem_version.rb
```diff
@@ -1,5 +1,5 @@
 module Xcodeproj
   # The version of the xcodeproj gem.
   #
-  VERSION = '0.27.0' unless defined? Xcodeproj::VERSION
+  VERSION = '0.27.1' unless defined? Xcodeproj::VERSION
```
segiddins: please don't bump the version -- that is done as part of the release process

segiddins:
\c @neonichu

segiddins:
This needs a changelog entry, tests, and a look into why CI is failing for us to consider the PR

segiddins:
(Additionally, I'm not convinced that just bumping the constants isn't the right solution)

carles:
Hi,

Sorry for the version bump, I’m not so familiar with the pull-request process.

And yes, bumping the constants would fix it for this particular type. However SDKs might change on each Xcode release so we’ll need to update the constants each time, particularly annoying during beta periods.  That’s why I thought that using the latest SDK found in the system made more sense, specially since I also found there was a method for that purpose.

Anyway, I only meant to raise this issue so that somebody with more experience might look into it.


Best,

carles

PR #308, opened 2015-09-26
Add `XCScheme#save!` to save scheme in place

AliSoftware:
When initialized from an existing `*.xcscheme` file, this allows to modify the existing schemes and save them more easily — as in such cases, the `save_as` (taking the path to the `xcodeproj` + scheme name as separate parameters) is not the most appropriate:

```ruby
s = Xcodeproj::XCScheme.new(some_path)
# … play with s, modify stuff
s.save!
```

### DIFF
--- a/lib/xcodeproj/scheme.rb
+++ b/lib/xcodeproj/scheme.rb
@@ -35,7 +35,6 @@ module Xcodeproj
 
         @scheme = @doc.elements['Scheme']
       else
-        @file_path = nil
         @doc = REXML::Document.new
         @doc.context[:attribute_quote] = :quote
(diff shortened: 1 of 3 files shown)


lib/xcodeproj/scheme.rb
```diff
@@ -29,11 +29,13 @@ class XCScheme
     #
     def initialize(file_path = nil)
       if file_path
+        @file_path = file_path
         @doc = REXML::Document.new(File.new(file_path))
         @doc.context[:attribute_quote] = :quote
 
         @scheme = @doc.elements['Scheme']
       else
+        @file_path = nil
```
segiddins: this is unecessary

AliSoftware: I knew you would say that, damn ^^

segiddins:
:+1: other than my one comment

PR #320, opened 2015-10-23
Adding .hxx to the list of recognized header file extensions

jasonvasquez:
I've run into a few cases with some third party open source libraries I'm wrapping in Cocoapods, where they use `.hxx` as a header extension in C++ projects.  I have .podspec `prepare_command`'s to actually rename all the files and update references to them, otherwise, we generate build rules for all the `.hxx` files.  That is obviously not optimal.  Xcode recognizes these as headers already, we just need to slot them in the right build phase.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@
   [Boris Bügling](https://github.com/neonichu)
   [Xcodeproj#318](https://github.com/CocoaPods/Xcodeproj/pull/318)
 
+* Added .hxx to the list of recognized header file extensions.  
+  [Jason Vasquez](https://github.com/jasonvasquez)
+  [Xcodeproj#320](https://github.com/CocoaPods/Xcodeproj/pull/320)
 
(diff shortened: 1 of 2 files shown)


orta:
seems :+1: to me

kylef:
Thanks for the pull request @jasonvasquez. Would you be able to add a short piece to the CHANGELOG so you get credit for this change please.

jasonvasquez:
done - thanks!

kylef:
Awesome. Until the next CocoaPod's release I'd advise you to stick to the prepare_command. Once it's out you can use it, but I would ensure you mark that your Pod would require X version of CocoaPods or newer to ensure that it isn't used on previous CocoaPod's installs.

For example:

```ruby
    spec.cocoapods_version = '>= 1.0.0'
```

https://guides.cocoapods.org/syntax/podspec.html#cocoapods_version

jasonvasquez:
Good idea - thanks!

PR #322, opened 2015-10-24
Adjust Xcodeproj::Workspace to keep the underlying REXML document and add basic <Group> support.

dparton:


lib/xcodeproj/workspace.rb
```diff
@@ -1,24 +1,48 @@
...
-    attr_reader :schemes
+    def file_references
+      return [] unless @document
+      @document.get_elements('/Workspace//FileRef').map do |node|
+        FileReference.from_node(node)
+      end
+    end
 
-    # @param  [Array] file_references @see file_references
+    # @return [Array<GroupReference>] the groups contained in the workspace
```
segiddins: this should still be `param`

lib/xcodeproj/workspace.rb
```diff
@@ -1,24 +1,48 @@
...
 
 module Xcodeproj
   # Provides support for generating, reading and serializing Xcode Workspace
   # documents.
   #
   class Workspace
-    # @return [Array<String>] the paths of the projects contained in the
+    # @return [REXML::Document] the parsed XML model for the workspace contents
+    attr_reader :document
+    attr_reader :schemes
```
segiddins: this attr needs to be documented

segiddins:
@AliSoftware could you review this, please?

lib/xcodeproj/workspace.rb
```diff
@@ -1,24 +1,48 @@
...
+        GroupReference.from_node(node)
+      end
+    end
+
+    # @param [REXML::Document] document @see document
+    # @param [Array<FileReference>] file_references additional projects to add
+    #
+    def initialize(document, *file_references)
       @schemes = {}
+      if !document || document.is_a?(REXML::Document)
```
AliSoftware: prefer `document.nil?` to `!document`

lib/xcodeproj/workspace.rb
```diff
@@ -1,24 +1,48 @@
...
+    end
 
-    # @param  [Array] file_references @see file_references
+    # @return [Array<GroupReference>] the groups contained in the workspace
     #
-    def initialize(*file_references)
-      @file_references = file_references.flatten
+    def group_references
+      return [] unless @document
+      @document.get_elements('/Workspace//Group').map do |node|
```
AliSoftware: Maybe we should _cache_ that (`@group_references ||= …`)?

lib/xcodeproj/workspace.rb
```diff
@@ -67,9 +89,36 @@ def self.from_s(xml, workspace_path = '')
     # @return [void]
     #
     def <<(projpath)
```
AliSoftware: Now that the parameter to this method can be either a `String` or a `FileReference`, please update the rdoc accordingly

lib/xcodeproj/workspace.rb
```diff
@@ -67,9 +89,36 @@ def self.from_s(xml, workspace_path = '')
...
+
+    # Adds a new group container to the workspace
+    # workspace.
+    #
+    # @param  [String] name The name of the group
+    #
+    # @return [Xcodeproj::Workspace::GroupReference]
+    # The added group reference
+    #
+    def add_group(name)
```
AliSoftware: Missing documentation about the optional block used by `yield`

lib/xcodeproj/workspace.rb
```diff
@@ -81,14 +130,36 @@ def <<(projpath)
...
     #
     def to_s
-      contents = file_references.map { |reference| file_reference_xml(reference) }
-      root_xml(contents.join(''))
+      contents = ''
+      stack = []
+      @document.root.each_recursive do |elem|
+        until stack.empty?
+          last = stack.last
+          if last == elem.parent
```
AliSoftware: Prefer using `break if last == elem.parent` then remove the `else`/`end` to avoid a useless level of indentation

lib/xcodeproj/workspace.rb
```diff
@@ -146,21 +217,36 @@ def root_xml(contents)
...
 <Workspace
    version = "1.0">
-   #{contents.strip}
+#{contents.rstrip}
 </Workspace>
       DOC
     end
 
-    # @return [String] The XML representation of a file reference.
+    # @return [String] The Xcode-specific XML formatting of an element start
```
AliSoftware: Convention is to put `@return` doc lines at the end after `@param` lines, not before

lib/xcodeproj/workspace/group_reference.rb
```diff
@@ -0,0 +1,64 @@
...
+      # @return [String] the name of the group
+      #
+      attr_reader :name
+
+      # @return [String] the type of reference to the project
+      #
+      # This can be of the following values:
+      # - absolute
+      # - group
+      # - container (only supported value)
```
AliSoftware: Should we throw in case we encounter a non-supported value, or just silently fail? (\cc @segiddins)

AliSoftware:
Please add an entry in CHANGELOG as well.
Otherwise, LGTM once the remarks have been addressed :+1:

lib/xcodeproj/workspace.rb
dparton: The idea of this change is to work against the underlying xml to allow the document structure to remain intact when working with workspaces.

lib/xcodeproj/workspace.rb
dparton: Are you saying the @return should still be @param?

lib/xcodeproj/workspace.rb
AliSoftware: Yeah gotcha, makes sense to ensure it's always in sync with the XML too.

lib/xcodeproj/workspace.rb
dparton: Apologies for my ignorance -- is this proper syntax for documenting a parameter with two allowed types? The yard @overload documentation seems cumbersome for something like this...

    # Adds a new path to the list of the of projects contained in the
    # workspace.
    # @param [String, Xcodeproj::Workspace::FileReference] path_or_reference
    #        A string or Xcode::Workspace::FileReference containing a path to an Xcode project
    #
    # @raise [ArgumentError] Raised if the input is neither a String nor a FileReference
    #
    # @return [void]
    #
    def <<(path_or_reference)

AliSoftware: Yes, YARD uses commas to separate the different possible types for a parameter.

For a quick cheat sheet, see http://yardoc.org/types.html

dparton:
I believe all the remarks are addressed - let me know if there's anything else that looks suspect or needs additional clarification/work

lib/xcodeproj/workspace/group_reference.rb
segiddins: Probably throw? But it's not super important

Issue #325, opened 2015-10-25
Add support for XCScheme environment variables


justinseanmartin:
I'd like to leverage the Xcodeproj library, but need support for reading/setting environment variables in a particular scheme. I'll open an associated PR with the supported behavior shortly.

segiddins:
Closed by https://github.com/CocoaPods/Xcodeproj/issues/326.

PR #326, opened 2015-10-25
Add support for scheme environment variables

justinseanmartin:
This resolves issue #325.

Aside from the stuff directly related to environment variables, I did some rubocop cleanup. After rebasing, the changes are mostly moot due to other changes to the .rubocop_cocoapods.yml. I also reran `rubocop --auto-gen-config`, which looks like it hasn't been run in a while. I can break these changes out to a separate PR, or remove them entirely if you'd prefer. LMK.

🌈 (after the fact)


lib/xcodeproj/scheme/environment_variables.rb
```diff
@@ -0,0 +1,117 @@
...
+  class XCScheme
+    VARIABLES_NODE = 'EnvironmentVariables'
+    VARIABLE_NODE = 'EnvironmentVariable'
+
+    # This class wraps the EnvironmentVariables node of a .xcscheme XML file. This
+    # is just a container of EnvironmentVariable objects. It can either appear on a
+    # LaunchAction or TestAction scheme group.
+    #
+    class EnvironmentVariables < XMLElementWrapper
+      # @param [REXML::Element,Array{EnvironmentVariable},Array{Hash{String => String}}] node_or_target
```
AliSoftware: I believe the YARD documentation syntax expect `Array<EnvironmentVariable>`, see http://yardoc.org/types.html for a quick cheat sheet.

lib/xcodeproj/scheme/environment_variables.rb
```diff
@@ -0,0 +1,117 @@
...
+      #   The 'EnvironmentVariables' XML node that this object represents.
```
AliSoftware: … or a list of `EnvironmentVariables`, or an array of `var => value` hashes.

BTW, why expect an `Array<Hash{String => String}>` instead of just a `Hash{String => String}` directly?

lib/xcodeproj/scheme/environment_variables.rb
```diff
@@ -0,0 +1,117 @@
...
+      #   If nil, it will create a default XML node to use.
+      #
+      def initialize(node_or_target = nil)
```
AliSoftware: Change the name of the parameter from `node_or_target` to something more suitable like `node_or_variables` for example

lib/xcodeproj/scheme/environment_variables.rb
```diff
@@ -0,0 +1,117 @@
...
+
+      # @return [Array{EnvironmentVariable}]
+      #   The key value pairs currently set in @xml_element
+      #
+      def all_variables
+        @all_variables ||= @xml_element.get_elements(VARIABLE_NODE).map { |variable| EnvironmentVariable.new(variable) }
+      end
+
+      # @param [EnvironmentVariable,Hash{String => String}] variable
+      #   The key value pairs currently set in @xml_element
```
AliSoftware: "currently set"? `add_variable` is to add new variables not currently in the node, right?

lib/xcodeproj/scheme/environment_variables.rb
```diff
@@ -0,0 +1,117 @@
...
+      #
+      # @return [Array{EnvironmentVariable}]
```
AliSoftware: `Array<EnvironmentVariable>`

lib/xcodeproj/scheme/environment_variables.rb
```diff
@@ -0,0 +1,117 @@
...
+        env_var = variable.is_a?(EnvironmentVariable) ? variable : EnvironmentVariable.new(variable)
+        all_variables.each { |existing_var| remove_variable(existing_var) if existing_var.key == env_var.key }
+        @xml_element.add_element(env_var.xml_element)
+        @all_variables << env_var
+      end
+
+      # @param [EnvironmentVariable,Hash{String => String}] variable
+      #   The key value pairs currently set in @xml_element
+      #
+      # @return [Array{EnvironmentVariable}]
```
AliSoftware: ditto

lib/xcodeproj/scheme/environment_variables.rb
```diff
@@ -0,0 +1,117 @@
...
+      #   The new set of environment variables after addition
+      #
+      def add_variable(variable)
+        env_var = variable.is_a?(EnvironmentVariable) ? variable : EnvironmentVariable.new(variable)
+        all_variables.each { |existing_var| remove_variable(existing_var) if existing_var.key == env_var.key }
+        @xml_element.add_element(env_var.xml_element)
+        @all_variables << env_var
+      end
+
+      # @param [EnvironmentVariable,Hash{String => String}] variable
```
AliSoftware: Couldn't this method also accept just a `String` containing only the name of the variable to remove?

lib/xcodeproj/scheme/environment_variables.rb
```diff
@@ -0,0 +1,117 @@
...
+      #   The key value pairs currently set in @xml_element
+      #
+      # @return [Array{EnvironmentVariable}]
+      #   The new set of environment variables after removal
+      #
+      def remove_variable(variable)
+        env_var = variable.is_a?(EnvironmentVariable) ? variable : all_variables.select { |var| var.key == variable[:key] }[0]
```
AliSoftware: Prefer `.find` instead of `select { … }[0]`

lib/xcodeproj/scheme/environment_variables.rb
```diff
@@ -0,0 +1,117 @@
...
+      end
+
+      # @return [Array{EnvironmentVariable}]
+      #   The key value pairs currently set in @xml_element
+      #
+      def all_variables
+        @all_variables ||= @xml_element.get_elements(VARIABLE_NODE).map { |variable| EnvironmentVariable.new(variable) }
+      end
+
+      # @param [EnvironmentVariable,Hash{String => String}] variable
```
AliSoftware: Add a description to this method as part of the doc, like "Adds the given variable to the EnvironmentVariables, or Replace it if it already exists".

Btw, maybe we should call that method `set_variable` instead of `add_variable` (because it doesn't blindly add but replace in case it exists already)?

AliSoftware:
Thanks for the PR, looks very promising!

Ok for me once my remarks have all been addressed.

lib/xcodeproj/scheme/environment_variables.rb
```diff
@@ -0,0 +1,117 @@
...
+        string_to_bool(@xml_element.attributes['isEnabled'])
+      end
+
+      def enabled=(enabled)
+        @xml_element.attributes['isEnabled'] = bool_to_string(enabled)
+      end
+
+      # @return [Hash{:key => String, :value => String, :enabled => Boolean}]
+      #   The environment variable XML node with attributes converted to Hash keys
+      def to_h
```
AliSoftware: The ruby standard name for this method is `to_hash`, not `to_h`

lib/xcodeproj/scheme/environment_variables.rb
```diff
@@ -0,0 +1,117 @@
...
+        create_xml_element_with_fallback(value, VARIABLE_NODE) do
+          raise "Must pass a Hash with 'key' and 'value'!" unless value.is_a?(Hash) && value.key?(:key) && value.key?(:value)
+
+          @xml_element.attributes['key'] = value[:key]
+          @xml_element.attributes['value'] = value[:value]
+          @xml_element.attributes['isEnabled'] = value.key?(:enabled) ? bool_to_string(value[:enabled]) : bool_to_string(true)
+        end
+      end
+
+      def key
```
AliSoftware: Please add YARD doc

lib/xcodeproj/scheme/environment_variables.rb
```diff
@@ -0,0 +1,117 @@
...
+        @xml_element.attributes['key']
+      end
+
+      def key=(key)
```
AliSoftware: Please add YARD doc

lib/xcodeproj/scheme/environment_variables.rb
```diff
@@ -0,0 +1,117 @@
...
+        @xml_element.attributes['key'] = key
+      end
+
+      def value
```
AliSoftware: Please add YARD doc

lib/xcodeproj/scheme/environment_variables.rb
```diff
@@ -0,0 +1,117 @@
...
+        @xml_element.attributes['value']
+      end
+
+      def value=(value)
```
AliSoftware: Please add YARD doc

lib/xcodeproj/scheme/environment_variables.rb
```diff
@@ -0,0 +1,117 @@
...
+        @xml_element.attributes['value'] = value
+      end
+
+      def enabled
```
AliSoftware: Please add YARD doc

lib/xcodeproj/scheme/environment_variables.rb
```diff
@@ -0,0 +1,117 @@
...
+        string_to_bool(@xml_element.attributes['isEnabled'])
+      end
+
+      def enabled=(enabled)
```
AliSoftware: Please add YARD doc

lib/xcodeproj/scheme/launch_action.rb
```diff
@@ -62,6 +62,28 @@ def buildable_product_runnable=(runnable)
         @xml_element.add_element(runnable.xml_element) if runnable
       end
 
+      # @return [EnvironmentVariables]
+      #        Returns the set of environment variables that are currently set at test runtime
+      #        NOTE: If this creates a new node, it will not automatically be added to the xml document
```
AliSoftware: Use `@note` for notes (+ add a description in the doc)

lib/xcodeproj/scheme/environment_variables.rb
justinseanmartin: I could change the Hash keys to be strings instead of symbols as the Hash keys if that is more standard across the project, doesn't make a difference to me, seems how the enforced style is `=>` and not `:` for Hashes.

Alternatives are:
 - I could also just not allow initialization via the Hash, but it seems like a nice convenience. 
 - I could remove the ability to set enabled state through it. 

I didn't particularly like either of those options however.

I agree that it is a bit confusing having something that is already almost like a Hash to be reflected not as a hash, because of the desire to reflect the enabled state. For now, I'll document this expectation here unless you think something else is more appropriate.

lib/xcodeproj/scheme/environment_variables.rb
justinseanmartin: Yep, you're right. First time for YARD, read through the documentation. Will fix this up all over.

lib/xcodeproj/scheme/environment_variables.rb
justinseanmartin: Done.

lib/xcodeproj/scheme/environment_variables.rb
justinseanmartin: Fixed.

lib/xcodeproj/scheme/environment_variables.rb
justinseanmartin: Fixed (all over).

lib/xcodeproj/scheme/environment_variables.rb
justinseanmartin: Yep, absolutely could, updated. Also, will add a short description.

lib/xcodeproj/scheme/environment_variables.rb
justinseanmartin: Fixed

lib/xcodeproj/scheme/environment_variables.rb
justinseanmartin: Can't use `set_` prefix per rubocop, but how about `upsert_variable`? I don't love it, but can't come up with anything better at the moment.

lib/xcodeproj/scheme/environment_variables.rb
AliSoftware: After reading the rest of the code and realizing that a EnvVariable is not just a key & value but also enabled, it seems OK to keep that. (Also please keep the keys as symbols like you did).

The only thing missing then is proper documentation to tell how this Hash is supposed to be organized and what key symbols are expected.

lib/xcodeproj/scheme/environment_variables.rb
justinseanmartin: Seems to be differing information. From what I've read, appears that `to_hash` is specifically if the object acts like a hash and can be converted with no loss of fidelity. I don't think that is the case here. I don't feel strongly though, so LMK if I'm just misunderstood.

lib/xcodeproj/scheme/environment_variables.rb
justinseanmartin: Done (and for all below)

lib/xcodeproj/scheme/launch_action.rb
justinseanmartin: I removed this, as it seems to be consistent with the other method behaviors, so appears to be expected.

lib/xcodeproj/scheme/environment_variables.rb
AliSoftware: What's `upsert`, a mix of `update` and `insert`? I don't like it either. Maybe `define_variable`?

lib/xcodeproj/scheme/environment_variables.rb
AliSoftware: Ah, actually you're right, didn't know about that. Let's keep `to_h` then.

lib/xcodeproj/scheme/environment_variables.rb
```diff
@@ -0,0 +1,149 @@
...
+      #
+      # @param [EnvironmentVariable,Hash{Symbol => String,Bool}] variable
+      #   The variable to set
+      # @note If initializing with an array of hashes, the hash should be of the form
+      #   {:key => <String>, :value => <String>(, :enabled => <Bool>)}.
+      # @see EnvironmentVariable#initialize
+      # @return [Array<EnvironmentVariable>]
+      #   The new set of environment variables after addition
+      #
+      def assign_variable(variable)
```
AliSoftware: I definitely prefer that `assign_variable`.

BTW, talking about `[]` and `[]=` it actually seem like a good addition (but still while keeping this `assign_variable` of course).

While `assign_variable` would take an `EnvironmentVariable, Hash{Symbol=>String,Bool}`, `[]=` would take the variable name as key and expect the variable content as value (and would assume `enabled=YES`, no way to specify `enabled` with that syntax). And `[]` would take the variable name as key and return the `EnvironmentVariable`.
That could provide a nice syntactic sugar alternative. What do you think?

justinseanmartin: Works for me. Should `[]` operate on enabled variables only as well?

AliSoftware: I don't think so, since it would return an `EnvironmentVariable` anyway, I'd say let it access all variables (enabled or disabled). The user could still check the variable state once retrieved.

justinseanmartin: Added.

lib/xcodeproj/scheme/environment_variables.rb
```diff
@@ -0,0 +1,171 @@
...
+    VARIABLES_NODE = 'EnvironmentVariables'
+    VARIABLE_NODE = 'EnvironmentVariable'
+
+    # This class wraps the EnvironmentVariables node of a .xcscheme XML file. This
+    # is just a container of EnvironmentVariable objects. It can either appear on a
+    # LaunchAction or TestAction scheme group.
+    #
+    class EnvironmentVariables < XMLElementWrapper
+      # @param [nil,REXML::Element,Array<EnvironmentVariable>,Array<Hash{Symbol => String,Bool}>] node_or_variables
+      #   The 'EnvironmentVariables' XML node that this object represents.
```
AliSoftware: The 'EnvironmentVariables' XML node, or the list of env variables, that this object represents.

AliSoftware:
Sorry not to have spotted that before, but could you fix the alignment/identation of all your YARD docs?

Convention is to align the text associated with `@param` so that it starts at the same column as the param type, [like this](https://github.com/CocoaPods/Xcodeproj/blob/master/lib/xcodeproj/workspace.rb#L44-L45)

```
# Some description
#
# @param [Type] paramName
#        This parameter description starts as the same indentation level as the type
#        and continues with the same alignment if it wraps on multiple lines
#
def foo(paramName)
  ...
end
```

lib/xcodeproj/scheme/environment_variables.rb
justinseanmartin: Fixed

justinseanmartin:
All feedback addressed and squashed the commits.

lib/xcodeproj/scheme/environment_variables.rb
```diff
@@ -0,0 +1,170 @@
...
+      #         Returns the matching environment variable for a specified key
+      #
+      def [](key)
+        all_variables.find { |var| var.key == key }
+      end
+
+      # Assigns a value for a specified key
+      #
+      # @param [String] key
+      #   The key to update in the environment variables
```
AliSoftware: Still misaligned ;)

lib/xcodeproj/scheme/environment_variables.rb
```diff
@@ -0,0 +1,170 @@
...
+      # @return [EnvironmentVariable] variable
+      #   The newly updated environment variable
+      #
+      def []=(key, value)
+        assign_variable(:key => key, :value => value)
+        self[key]
+      end
+
+      # @return Array<Hash{Symbol => String,Bool}>
+      #   The current environment variables represented as an array
```
AliSoftware: Still misaligned :wink:

justinseanmartin: And missing `[]` around the return type! Fixed in next commit.

lib/xcodeproj/scheme/environment_variables.rb
justinseanmartin: Fixed.

AliSoftware:
LGTM now :+1: Will merge in a few once Travis is OK.

Nice work @justinseanmartin :tada:

justinseanmartin:
Thanks for the prompt review feedback and dealing with a first time project contributor :smiley:

AliSoftware:
Then even more awesome work for a first-time contributor then! :+1:

tirodkar:
I don't think this is pushed out to the latest release yet. This would be a great enhancement to have. Is there an ETA for the next release?

segiddins:
@tirodkar there's no fixed release schedule, but this will ship with 1.0

PR #335, opened 2015-10-30
Escape XML entities in workspace file reference nodes

cjwirth:
Closes https://github.com/CocoaPods/CocoaPods/issues/4446.

This is my first time contributing to the base project, so if this isn't where this logic should go, let me know (or if I'm just simply doing it wrong).

This is meant to fix the issue ["Can't use '&' in Xcode project's name."](https://github.com/CocoaPods/CocoaPods/issues/4446) in the main repository. 

I tried doing the same thing with < and >, and it seemed like HTML entities need to be escaped, so I added it in. That's all.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,9 @@
   [Samuel Giddins](https://github.com/segiddins)
   [CocoaPods#3652](https://github.com/CocoaPods/CocoaPods/issues/3652)
 
+* Escape XML entities in project names when writing workspace
+  [Caesar Wirth](https://github.com/cjwirth)
+  [CocoaPods#4446](https://github.com/CocoaPods/CocoaPods/issues/4446)
 
 ## 0.28.2 (2015-10-09)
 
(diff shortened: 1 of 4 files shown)


cjwirth:
Whoops nevermind, tests failing

cjwirth:
OK, sorry for jumping the gun on that last one. It ran fine, but the tests failed.

Interestingly, without this fix, the tests still pass, but actually running `pod install` will fail. Now I've actually got a better fix for you, with the tests passing.

The problem was that the Attributes that it was actually writing out was the `attribute.value`. According to the [documentation](http://www.germane-software.com/software/XML/rexml/doc/classes/REXML/Attribute.html#M000452), you will have an unnormalized version (ie, &amp; will be turned into &). However, the normal string representation will still be normalized (ie. &amp; will still be &amp;.)

I think the tests were passing because it was actually comparing the 2 strings, rather than the 2 values. 

Originally I was going to put in `.to_s` to be more explicit, but the tests gave me a warning, so I took it out.

segiddins:
Thanks so much for this! Do you think you could add a simple test and a changelog entry? Thanks!

cjwirth:
The workspace test just had a "formats XML correctly" test, so I added some characters to escape in there. Hope that's what you were looking for?

segiddins:
👍🏻 ace

PR #336, opened 2015-10-31
[AbstractTarget] Add a helper to set the deployment target

segiddins:
Needed for https://github.com/CocoaPods/CocoaPods/pull/4317.

\c @neonichu @kylef

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,10 @@
   [David Parton](https://github.com/dparton)
   [#322](https://github.com/CocoaPods/Xcodeproj/pull/322)
 
+# Added a helper to set the deployment target on a target based on its
+  platform.  
+  [Samuel Giddins](https://github.com/segiddins)
+
 ##### Bug Fixes
 
 * Avoid duplicating settings with multiple values in common when merging.  
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,7 @@
   [David Parton](https://github.com/dparton)
   [#322](https://github.com/CocoaPods/Xcodeproj/pull/322)
 
-# Added a helper to set the deployment target on a target based on its
+* Added a helper to set the deployment target on a target based on its
   platform.  
   [Samuel Giddins](https://github.com/segiddins)
 
--- a/lib/xcodeproj/project/object/native_target.rb
+++ b/lib/xcodeproj/project/object/native_target.rb
@@ -108,15 +108,34 @@ module Xcodeproj
           sdk.scan(/[0-9.]+/).first
         end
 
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -25,6 +25,10 @@
   [David Parton](https://github.com/dparton)
   [#322](https://github.com/CocoaPods/Xcodeproj/pull/322)
 
+# Added a helper to set the deployment target on a target based on its
```
kylef: Should be a list item instead of a heading.

lib/xcodeproj/project/object/native_target.rb
```diff
@@ -108,15 +108,34 @@ def sdk_version
           sdk.scan(/[0-9.]+/).first
         end
 
+        # @visibility private
```
kylef: Shouldn't this just be `private` outside of a comment?

segiddins: No, it's not possible to actually make constants private.

AliSoftware:
LGTM :shipit:

lib/xcodeproj/project/object/native_target.rb
```diff
@@ -108,15 +108,34 @@ def sdk_version
...
         # @return [String] the deployment target of the target according to its
         #         platform.
         #
         def deployment_target
-          case platform_name
-          when :ios then common_resolved_build_setting('IPHONEOS_DEPLOYMENT_TARGET')
-          when :osx then common_resolved_build_setting('MACOSX_DEPLOYMENT_TARGET')
-          when :tvos then common_resolved_build_setting('TVOS_DEPLOYMENT_TARGET')
-          when :watchos then common_resolved_build_setting('WATCHOS_DEPLOYMENT_TARGET')
+          setting = DEPLOYMENT_TARGET_SETTING_BY_PLATFORM_NAME[platform_name]
```
floere: `platform_name` can be `nil` in case `sdk` is falsy or it does not include any of the options in the `platform_name` method. Perhaps we should `raise` – or inform the user if that is the case (either in here or in the `platform_name` method).

segiddins: I think it's intentional to not have `nil` cause things to blow up here -- but I will add a `return unless`

floere: Ah, it should blow up somewhere else? I am wondering if it should blow up here or in `platform_name`.

segiddins: no, it shouldnt blow up at all and just return `nil` :P

segiddins:
:shipit:

PR #338, opened 2015-11-09
Track .def files as headers

dnkoutso:
@segiddins I apologize for this inconvenience.

My original PR here tracked these files as sources but it was a mistake. `.def` files are meant to be used as headers.

By adding them as source files we are now receiving warnings when xcode is trying to compile them.

Please accept this PR and I will revert the source file tracking from CocoaPods and update the CHANGELOG accordingly.


segiddins:
👍🏻

Issue #343, opened 2015-12-07
Method for setting up a clean project


jcampbell05:
Currently Xcodeproj's method for creating a project makes a few assumptions which is great for CocoaPods but not so much for other gems.

In Xcake I made a subclass where I could setup the Project my own custom way:
https://github.com/jcampbell05/xcake/blob/master/lib/xcake/xcode/project.rb

When I was creating a project where it wasn't setting the object version so I had to manually make it return latest all the time.

It also added two configurations "Debug" and "Release",

Xcake needed for these to not be there so it could keep track of what it had added (The User may for some reason not like these names in which case Xcake would have has to remove them). So I only add configurations the user specifies (or debug or release as defaults if they don't but the gem still retains full control), my method in my subclass only setups up a blank project, nothing more, nothing less.

I realise that without these configurations the Xcode project can't open, so maybe there is some way when we save we could generate some default configurations for it to open :)

segiddins:
You can skip Xcodeproj's initialization if you want.

`- (Project) initialize(path, skip_initialization = false, object_version = Constants::DEFAULT_OBJECT_VERSION)`

I believe that initialization is the same as what Xcode will do by default, though, which is why Xcodeproj (by default) does it.

jcampbell05:
@segiddins  Initially that is what I tried however:

1. The Object Version wasn't set in that particular branch of the project setup, so it broke.
2. I needed a path that deviated from the standard Xcode method and created a barebones project for my library (Just the project, no targets and no configurations).

I can create an example for 1. and we can probably patch it (It may even be fixed now, I'll check).

But for 2. I had to create an extra method in my gem above that just setup the project without any default configurations. The current methods are good for CocoaPods or Users wanting quick and easy defaults like Xcode does it, but it gets in the way if you want full control.

It could be an extra method; So we have "initialize_from_scratch", "initialize_from_file". The current "from scratch" behavior could be "initialize_from_xcode_defaults" as although tecnically its creating a new project it didn't match my concept of "from scratch" it was more of a "ready meal" :) and what I mention above could be the new "initialize_from_scratch" (Ofc we have to take API compatibility into account)

segiddins:
Right, if you're skipping initialization, you'll need to set the object version yourself. Another alternative is to just subclass and override `initialize_from_scratch`, since the goal of Xcodeproj is _explicitly_ to set things up exactly as Xcode does.

jcampbell05:
[...]

PR #344, opened 2015-12-07
Additional documentation when creating project from scratch.

jcampbell05:
Adds additional documentation for #343, which explains that Xcodeproj is emulating Xcode behaviour and for those, like myself, who just wanted a true from scratch project then we must subclass and override `initialize_from_scratch` and set the object version.


orta:
Nice work

jcampbell05:
@orta  Cheers ! :)

segiddins:
🚀 thanks!

PR #351, opened 2016-01-05
Set TARGETED_DEVICE_FAMILY to include iPhone and iPad device families…

arifken:
We found that the TARGETED_DEVICE_FAMILY is not set correctly for resource bundles configured in a podspec. The application using the resource bundle from the pod did not compile assets for iPad because this value was defaulted to "1" (iPhone only). Now, it correctly configures the pod resource bundle target to include both iPhone and iPad

### DIFF
--- a/lib/xcodeproj/constants.rb
+++ b/lib/xcodeproj/constants.rb
@@ -270,6 +270,7 @@ module Xcodeproj
       }.freeze,
       [:ios, :bundle] => {
         'SDKROOT'                           => 'iphoneos',
+        'TARGETED_DEVICE_FAMILY'            => '1,2',
       }.freeze,
       [:osx, :bundle] => {
         'CODE_SIGN_IDENTITY'                => '-',


segiddins:
I think this should be set in CocoaPods, as the Xcode default does seem to be `1` and Xcodeproj attempts to mimix what Xcode does.

arifken:
Ah ok, where do you think we should apply this setting in the CocoaPods project?

segiddins:
https://github.com/CocoaPods/CocoaPods/blob/e28e4c1cdb721b195dcdeff209573c31fa49cd63/lib/cocoapods/installer/target_installer/pod_target_installer.rb#L176, but you'll have to scope it to only be when the platform name is `:ios`. You'll also want to add test coverage and a changelog entry. Thanks!

arifken:
Ok thanks. Also worth mentioning, I think that Xcode 7 automatically sets the TARGETED_DEVICE_FAMILY to 1,2 when you create a new Bundle target in Xcode by (clicking the + in the project settings). That's how we figured out why we were only having issues with iPad images in CocoaPods bundles and not bundles created in Xcode.

arifken:
@segiddins I created https://github.com/CocoaPods/CocoaPods/pull/4734 for this issue, so I'm going to close this PR. Thanks!

PR #358, opened 2016-02-16
Fix documentation for move and move_from methods on object_list

BergQuester:
The documentation for move and move_from methods on object_list had a grammatical error. Additionally, the description for move was incorrect.


segiddins:
Thanks!

PR #367, opened 2016-04-20
Added support for saving schemes to workspaces.

informatimago:
When the scheme is saved in a workspace, and makes reference to targets
in projects that are not stored in the same directory as the
workspace, the container: fields must contain a path to the project
relative to the directory where the workspace is stored.  For this, we
need to set the bundlePath (and give the scheme name and shared flag
to build the scheme file name at the same time) before we add the
targets, so the XML is created correctly.  Unfortunately the XML code
is generated when we add objects, and the Xcodeproj objects are thrown
away immediately, keeping only the REXML::Element, so it's not
possible to do it when we save, and it is impossible to save different
workspace (in different directories) from the same scheme object!

           scheme = Xcodeproj::XCScheme.new
           scheme.setBundlePathAndName dirPath+"MyWorkspace.xcworkspace","All",true
           scheme.add_build_target targets[0]
           scheme.add_build_target targets[1]
           # …
           scheme.save!

With more time, this would have to be rewritten entirely, so that we may add the targets, and set where to save it, in any order; ie. the XML generation should not occur when we add the targets and other elements, but when we save the scheme file, after having filled all the required data including where to save it. Notably, if we save the same Xcodeproj::XCScheme instance to different workspaces (in different directories) then the generated XML has to be different (different relative paths to the projects).


segiddins:
@AliSoftware would you like to review this?

segiddins:
@informatimago it appears things are failing on CI?

informatimago:
Yes I've seen; it looked to me unrelated to my changes. But I haven't considered tests at all even locally, and surely I'd have to add a few.  I'll have a look at it in the week end, I won't be able earlier. 

-- 
__Pascal Bourguignon__

lib/xcodeproj/scheme/buildable_reference.rb
```diff
@@ -44,7 +47,7 @@ def target_referenced_container
       end
 
       # Set the BlueprintIdentifier (target.uuid), BlueprintName (target.name)
-      #     and TerefencedContainer (URI pointing to target's projet) all at once
+      #     and RerefencedContainer (URI pointing to target's projet) all at once
```
mrackwitz: I guess that should be `ReferencedContainer` instead?

AliSoftware:
Woods how come I missed that PR? Sorry, will try and review it asap!

AliSoftware:
Alright, after a quick review:

* Please **fix Unit Tests (specs)**. It seems like you didn't update any of them (hence Travis-CI failing and the test not passing)
  * it's important to check that your changes don't introduce regressions and work both when saving the scheme in a xcworkspace or in a project, shared or not, with the `bundlePath` changed between saves, …
  * ➡️ You can run `rake spec` locally to run the unit tests and see if you fixed all the existing ones and if the ones you added also pass.
* Could you also **please fix rubocop** errors? Rubocop ensures we have some **code style consistency** in our Ruby code. Especially I see that:
  * you use `if … ; then` constructs where we prefer `if …` (no need for `then`)
  * you should add a space after your commas.
  * ➡️ You can use `rake rubocop` to check your code style and fix convention errors quickly.


_We won't be able to accept this PR until those changes are made so that we have the assurance that it won't break every use case covered by the specs._

Thanks. 👍

patchymcpatchface:
Hi!  

(Thank you for the instructions about rake spec and rubocop, I didn't know about that yet).

I've updated from the HEAD and corrected all the failing tests and rubocop issues (why doesn't rubocop correct them itself?).

Could you please consider pulling my changes again?

Thanks.

informatimago:
Hi!

(Thank you for the instructions about rake spec and rubocop, I didn't know about that yet).

I've updated from the HEAD and corrected all the failing tests and rubocop issues (why doesn't rubocop correct them itself?).

Could you please consider pulling my changes again?

Thanks.

AliSoftware:
Will try to take a look during the week

AliSoftware:
I quite don't like that we now have to pass the `scheme` to each and every method and object needed to manipulate the xcscheme.

You usually know the scheme and especially the relative path ahead of the time when you create the scheme, so adding that constraint just for the time you don't seems really too much change and constraint for the API imho.

For the rare kind of situation where you want to duplicate an existing xcscheme and just change the relative paths used everywhere, you could just visit every node and replace the old path with the new one, right?

Because pulling that `scheme` parameter everywhere only to address this use case is not worth it imho.
Redesigning the `XCScheme` ruby class to store the values in local objects first, and only create the backing XML representation on `save()` would probably be a better solution in the long run.

informatimago:
I entirely agree: it's a Q&D patch, and I would like to refactor the xcscheme module as you described, but for lack of time I had to make it like this.  But I consider it temporary, and since I will want to add other features to edit the parameters of the schemes easily (for now I use a sed phase to insert the parameters I need), I will eventually have to do this refactoring, creating the XML only on save.

AliSoftware:
Ok I understand the idea.

But I think we'll still wait for someone to have time to refactor properly in this case, because merging this implementation in the meantime would mean change the API and bump th major version of the Gem… only to restore the old API (with possibly only a new, optional parameter to the save method to give the relative path) once the refactoring is done.

So merging this patch changing the API then merge a future refactor would mean bumping the major version twice and make people using the Xcodeproj API change their code twice (back and forth) — with a lot of people not even needing this feature — while having the real refactoring would only mean a minor version bump — and no changes for current consumers of the API.

Sorry you went to all this trouble for this PR — in any way don't feel discouraged if we don't merge it, the effort was greatly appreciated and the discussion insightful to understand your needs will definitely help in the future refactoring of this API.

PR #375, opened 2016-05-25
Fix json gem with Ruby 2.2

chrisballinger:
https://github.com/CocoaPods/Rainforest/issues/63


segiddins:
You also need to run `bundle install`

PR #386, opened 2016-06-13
add targets json output

dacaiguoguo:
add targets json output. can use to filemerge compare to targets diffs

### DIFF
--- a/lib/xcodeproj/command/show.rb
+++ b/lib/xcodeproj/command/show.rb
@@ -46,6 +46,19 @@ module Xcodeproj
         pretty_print = xcodeproj.pretty_print
         sections = []
         pretty_print.each do |key, value|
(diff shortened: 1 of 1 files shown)


segiddins:
Sorry, but I'm not really sure this is a feature we need to add, especially since we already emit YAML there

orta:
Aye, agreed

dacaiguoguo:
xcodeproj target-diff only output --- []
my shell script ：
```
#!/bin/sh
# ./findtargetdiff.sh LABC.xcodeproj LABC LABCEnterPrise
# ./findtargetdiff.sh LABC.xcodeproj LABC LABCBack
xcodeproj sort $1 > /dev/null
xcodeproj show $1 > /dev/null
opendiff $2.json $3.json
```
 json is good

PR #396, opened 2016-07-17
Support for identifying hosts of embedded targets from sub-projects

benasher44:
This is the first step to fixing CocoaPods/CocoaPods#5622

### DIFF
--- a/lib/xcodeproj/project/object/target_dependency.rb
+++ b/lib/xcodeproj/project/object/target_dependency.rb
@@ -47,6 +47,7 @@ module Xcodeproj
         def native_target_uuid
           return target.uuid if target
           return target_proxy.remote_global_id_string if target_proxy
+          raise 'Expected target or target_proxy, from which to fetch a uuid'
         end
 
         # @note This override is necessary because Xcode allows for circular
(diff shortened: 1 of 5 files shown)


benasher44:
Looks like travis failed installing json.

benasher44:
Same issue.

lib/xcodeproj/project/object/target_dependency.rb
```diff
@@ -39,6 +39,16 @@ def display_name
...
         end
 
+        # @return [String] uuid of the target, if the dependency
+        #         is a native target, otherwise the uuid of the
+        #         target in the sub-project if the dependency is
+        #         a target proxy
+        #
+        def native_target_uuid
+          return target.uuid if target
+          return target_proxy.remote_global_id_string if target_proxy
```
segiddins: maybe raise otherwise to ensure this doesn't return nil?

segiddins:
👍

lib/xcodeproj/project/object/target_dependency.rb
benasher44: This is actually fixed in the line below, but it didn't close this comment since I didn't change these lines :)

segiddins:
@benasher44 this should pass with a rebase

benasher44:
This is ready on my end! Once this is merged, I can update CocoaPods/CocoaPods#5637

PR #415, opened 2016-09-09
[FFI] Default to overriding chdir

DanToml:
Because running Xcode via `open3` is so slow, default to the `chdir` override,  but allow users to opt for `open3` via an env var.

### DIFF
--- a/lib/xcodeproj/plist/ffi.rb
+++ b/lib/xcodeproj/plist/ffi.rb
@@ -1,5 +1,35 @@
 module Xcodeproj
   module Plist
+    module Extensions
+      # Since Xcode 8 beta 4, calling `PBXProject.projectWithFile` does break subsequent calls to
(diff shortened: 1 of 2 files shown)


DanToml:
There seems to be a Bundler regression in 1.13.0 on Ruby 2.1 https://travis-ci.org/CocoaPods/Xcodeproj/jobs/158887964 - Trying pinning to 1.12.5 for now.

lib/xcodeproj/plist/ffi.rb
```diff
@@ -1,5 +1,35 @@
 module Xcodeproj
   module Plist
+    module Extensions
+      # Since Xcode 8 beta 4, calling `PBXProject.projectWithFile` does break subsequent calls to
```
orta: > does break subsequent

breaks

DanToml:
@orta updated.

orta:
👍 looks good to go from my side

Issue #221, opened 2014-12-08
FileReferencesFactory#new_reference doesn't handle `xcdatamodeld` files properly


gfontenot:
Been talking about this in thoughtbot/liftoff#192 for a while, but I wanted to open an issue here too, since I'm more and more convinced that this is just a bug in Xcodeproj.

When using `FileReferencesFactory#new_reference`, the model is added incorrectly. Here is the diff of a project that had the model added with Xcodeproj, then had the model removed from the project and added back in manually:

``` diff
diff --git a/Testing.xcodeproj/project.pbxproj b/Testing.xcodeproj/project.pbxproj
index 1ff1901..6a46448 100644
--- a/Testing.xcodeproj/project.pbxproj
+++ b/Testing.xcodeproj/project.pbxproj
@@ -9,7 +9,7 @@
 /* Begin PBXBuildFile section */
        2494C90A2A249847CFF57868 /* Foundation.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 48B899A3BD55ABA1DB000BB2 /* Foundation.framework */; };
        9EA02258BC08067F87A19B3A /* Settings.bundle in Resources */ = {isa = PBXBuildFile; fileRef = 50DF5B4A67681D5EAAA824CB /* Settings.bundle */; };
-       FD1BDEAC9087955B6F09B247 /* Model.xcdatamodeld in Sources */ = {isa = PBXBuildFile; fileRef = 389739C8263048FE9D4B250F /* Model.xcdatamodeld */; };
+       F8931E481A3224F900E304D3 /* Model.xcdatamodeld in Sources */ = {isa = PBXBuildFile; fileRef = F8931E461A3224F900E304D3 /* Model.xcdatamodeld */; };
 /* End PBXBuildFile section */

 /* Begin PBXContainerItemProxy section */
@@ -28,6 +28,7 @@
        50DF5B4A67681D5EAAA824CB /* Settings.bundle */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = "wrapper.plug-in"; path = Settings.bundle; sourceTree = "<group>"; };
        7CCCC30571AF6D8B92CD2BB9 /* UnitTests-Info.plist */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = text.plist.xml; path = "UnitTests-Info.plist"; sourceTree = "<group>"; };
[...]

PR #427, opened 2016-10-10
#221 Fixed handling of xcdatamodeld packages in subfolders

Eldorado234:
This fixes issue #221.

The PR #291  from @adamhrz unfortunately only works when the model is one folder deep in the hierarchy. As soon as it is located on a deeper level (e.g. `Source/CoreData/Model.xcdatamodeld`), PR #291 leads to a crash.

The underlying issue was that the `path` is relative to the `group` but was handled as it would be absolute. I just joined the absolute/real path of the `group` with the `path` to create the correct `real_path`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@
   [Danielle Tomlinson](https://github.com/dantoml)
   [#421](https://github.com/CocoaPods/Xcodeproj/pull/421)
 
+* Fixed handling of xcdatamodeld packages in subfolders.
+  [Simon Seyer](https://github.com/Eldorado234)
+  [#427](https://github.com/CocoaPods/Xcodeproj/pull/427)
 
 ## 1.3.1 (2016-09-10)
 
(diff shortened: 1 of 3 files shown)


segiddins:
Could you add test coverage and a CHANGELOG entry for this fix? Thanks!

Eldorado234:
@segiddins done.

 I've tested the test case with and without my changes. With my changes it succeeded, without it failed.

I'm not really experienced in Ruby so please let me know if there is anything I could improve. Thanks!

CHANGELOG.md
```diff
@@ -12,6 +12,9 @@
   [Danielle Tomlinson](https://github.com/dantoml)
   [#421](https://github.com/CocoaPods/Xcodeproj/pull/421)
 
+* Fixed handling of xcdatamodeld packages in subfolders.
```
DanToml: Could you please add two spaces to the end of this? i.e `* Fixed handling of xcdatamodeld packages in subfolders.  `

Eldorado234:
Finally the travis build finished (they had some [incident](https://www.traviscistatus.com/) that caused the delay)

segiddins:
👍

PR #428, opened 2016-10-12
Add product type constant for XPC services

IngmarStein:
XPC services might need to be added to CocoaPods's list of product types that require a host target. Otherwise, the Swift runtime is included multiple times, for example.


IngmarStein:
Yep. I'll try to figure out if that applies to all XPC services or if LD_RUNPATH_SEARCH_PATH needs to include the hosting application to avoid duplication (in which case I would not push the general PR).

Issue #409, opened 2016-09-02
product_ref_group is nil when loading a project


jeduden:
When loading a project from file the product_ref_group is nil even if the file contains the Products group. 
Initialize_from_scratch initializes the variable: https://github.com/CocoaPods/Xcodeproj/blob/master/lib/xcodeproj/project.rb#L184

However, initialize_from_file does not:
https://github.com/CocoaPods/Xcodeproj/blob/master/lib/xcodeproj/project.rb#L198

I would like to contribute a fix, however I am not sure where i can best add a test to reproduce the issue.

PR #429, opened 2016-10-13
Fix regression resulting in projects existing product refs being altered

rasharab:
Buddybuild uses xcodeproj against thousands of xcode projects, and there are a few we’ve noticed issues with as a result of this commit:

36ad554b Fixes issue CocoaPods/Xcodeproj#409

The problem is that xcodeproj was actually significantly altering the structure of a project if it already has a product_ref_group defined and it wasn’t a “Products” main group.

This would resultantly lead to fatal failures when we attempt to build the project.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,9 +8,10 @@
 
 ##### Bug Fixes
 
-* Fix regression resulting in projecs existing product ref groups being altered
+* Fix regression resulting in projects existing product ref groups being altered  
   [Rashin Arab](https://github.com/rasharab)
   [#429](https://github.com/CocoaPods/Xcodeproj/pull/429)
+
 
 ## 1.3.2 (2016-10-10)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,9 @@
 
 ##### Bug Fixes
 
-* None.  
-
+* Fix regression resulting in projecs existing product ref groups being altered
+  [Rashin Arab](https://github.com/rasharab)
+  [#429](https://github.com/CocoaPods/Xcodeproj/pull/429)
 
 ## 1.3.2 (2016-10-10)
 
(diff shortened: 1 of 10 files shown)


alloy:
Thanks! Could you please add a test and a CHANGELOG entry?

orta:
Ah, nice one - I've seen a few error reports around this https://github.com/CocoaPods/CocoaPods/issues/5847 I think is one still open - thanks!

rasharab:
Will do guys. :)
-Rashin

rasharab:
I've pushed the tests.
Waiting an interminable amount of time for the checks to go through...
Also, when this goes through, can we get a quick release. Our customers are being affected by this?

lib/xcodeproj/project.rb
```diff
@@ -216,7 +216,11 @@ def initialize_from_file
       if object_version.to_i > Constants::LAST_KNOWN_OBJECT_VERSION
         raise '[Xcodeproj] Unknown object version.'
       end
-      root_object.product_ref_group = root_object.main_group['Products'] || root_object.main_group.new_group('Products')
+
+      # Projects can have product_ref_groups that are not listed in the main_groups["Products"]
+      if root_object.product_ref_group.nil?
+        root_object.product_ref_group = root_object.main_group['Products'] || root_object.main_group.new_group('Products')
```
segiddins: maybe `root_object.product_ref_group ||= root_object.main_group['Products'] || root_object.main_group.new_group('Products')` ?

rasharab: That works as well, it's a bit more implicit, but I don't really have an issue either way.
I'll push a change.

rasharab: Done. :)
-Rashin

CHANGELOG.md
```diff
@@ -8,8 +8,9 @@
 
 ##### Bug Fixes
 
-* None.  
-
+* Fix regression resulting in projecs existing product ref groups being altered
```
DanToml: Please add two spaces to the end of the changelog entry (i.e `* Fix regression resulting in projecs existing product ref groups being altered  `)

rasharab: Done.

segiddins:
Thanks!

PR #430, opened 2016-10-15
[Plist] Migrate to purely native ASCII plist parsing

segiddins:
Uses https://github.com/dantoml/ascii_plist, which is close to a 0.1.0 release and will be transferred to the CocoaPods org really soon

This is what I promised I was working on at NSSpain 😉 

\c @DanToml @alloy @neonichu


alloy:
That implementation is pretty damn sweet 👏

samdmarshall:
this is pretty rad! 🎉

benasher44:
🎉🎉

segiddins:
This now does a lot more to match how Xcode deals with serializing build settings, review of that code would be appreciate

lib/xcodeproj/plist.rb
```diff
@@ -1,10 +1,10 @@
+autoload :Nanaimo, 'nanaimo'
+autoload :CFPropertyList, 'cfpropertylist'
```
DanToml: I'd think we should just require these tbh

lib/xcodeproj/project/object/build_configuration.rb
```diff
@@ -74,6 +80,52 @@ def sorted_build_settings
...
+          OTHER_CPLUSPLUSFLAGS
+          OTHER_LDFLAGS
+          REZ_SEARCH_PATHS
+          SECTORDER_FLAGS
+          WARNING_CFLAGS
+          WARNING_LDFLAGS
+        ).freeze
+        private_constant :ARRAY_SETTINGS
+
+        def normalize_array_settings(settings)
```
DanToml: The normalisation methods should probably be private because they don't really matter to users of Xcodeproj (as it's handled internally.)

spec/project_spec.rb
```diff
@@ -237,24 +237,33 @@ module ProjectSpecs
...
-      it 'escapes non ASCII characters in the project' do
-        Plist::FFI.stubs(:ruby_hash_write_xcode).returns(false)
+      it 'can open a project and save it without causing a diff' do
+        @project.mark_dirty!
+        @project.save(@tmp_path)
+        project_file = (temporary_directory + 'Pods.xcodeproj/project.pbxproj')
+
+        begin
+          File.read(project_file).should == File.read(@path + 'project.pbxproj')
+        rescue Bacon::Error => e
```
DanToml: It would be nice if we had a reusable marcher for diffing files

spec/project_spec.rb
```diff
@@ -237,24 +237,33 @@ module ProjectSpecs
...
+      it 'saves non ASCII characters in the project' do
         file_ref = @project.new_file('わくわく')
         file_ref.name = 'わくわく'
         file_ref = @project.new_file('Cédric')
         file_ref.name = 'Cédric'
         @project.save(@tmp_path)
         contents = File.read(@tmp_path + 'project.pbxproj')
-        contents.should.not.include('わくわく')
-        contents.should.include('&#12431;&#12367;&#12431;&#12367;')
-        contents.should.not.include('Cédric')
```
DanToml: Doesn't non ascii stuff break app code?

lib/xcodeproj/project/object/build_configuration.rb
segiddins: already are after a `private`

spec/project_spec.rb
segiddins: That would be called `rspec` :trollface:

spec/project_spec.rb
segiddins: this test was testing the XML serialization. I believe it works with next step plists? Because that's how Xcode will serialize them, best I could tell from my experimentation

lib/xcodeproj/project/object/build_configuration.rb
DanToml: Whoops missed that only at the diff 👍

lib/xcodeproj/plist.rb
segiddins: I don't like requiring everything upfront -- that would mean just `require`ing Xcodeproj would cause a large amount of potentially irrelevant code to be loaded upfront

spec/project_spec.rb
DanToml: Don't tempt me to file that PR more than I already am 😂

lib/xcodeproj/plist.rb
DanToml: Eh, Id say the overhead is minimal enough that it would be preferable to having problems when you come to loading it later, and there are very few uses of Xcodeproj that don't require some form of reading/writing. But fair enough ¯\_(ツ)_/¯ 👍

spec/project_spec.rb
DanToml: Cool 👍

DanToml:
🎉

alloy:
Awesome work 💯 Also love that name.

PR #432, opened 2016-10-17
Associated changelog entry to the correct version

Eldorado234:
Probably when merging the PR the changelog entry got associated with `1.3.2` in the changelog even though it is not released so far.


orta:
👍

PR #443, opened 2016-11-06
Add missing file extensions for target types

keith:
Currently if xcodeproj doesn't have an entry in this hash for a specific
target type, it generates a target like `Foo.` where it is missing the
extension. This change adds the extensions for a few more target types.


keith:
Thanks!

PR #456, opened 2016-12-09
add more HEADER_EXTENSIONS

king6cong:
https://github.com/CocoaPods/CocoaPods/pull/6285

PR moved here as advised by @benasher44

### DIFF
--- a/lib/xcodeproj/constants.rb
+++ b/lib/xcodeproj/constants.rb
@@ -355,6 +355,6 @@ module Xcodeproj
 
     # @return [Array] The extensions which are associated with header files.
     #
-    HEADER_FILES_EXTENSIONS = %w(.h .hh .hpp .ipp .tpp .hxx .def).freeze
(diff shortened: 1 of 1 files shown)


benasher44:
@king6cong this generally looks good! Can you provide some more context around the .msg-type file? I can't seem to find documentation on this one. Thanks!

benasher44:
@king6cong I went ahead and tried adding these kinds of files to a framework target in Xcode, and Xcode didn't properly add them to the header build phase. Because Xcode doesn't know what to do with these kinds of files by default (ideally the files should get added to the header build phase), a custom build phase would need to be added to handle these files types. Therefore, this PR doesn't seem right for Xcodeproj: Xcode itself doesn't recognize these file types as headers, and Xcodeproj should be left as is to reflect that.

I do appreciate you putting together this PR though. I definitely learned a few things today!

PR #461, opened 2017-01-17
[Plist] Correct auto-load for CFPropertyList

segiddins:
The file is actually capitalized, so this will now work on case-sensitive file systems


DanToml:
Seems to have broken them?

segiddins:
Hmmm for some reason it appeared capitalized on my local machine. I'll look into it.

PR #463, opened 2017-01-27
Switch from unmaintained colored to maintained colored2

andrew:
The colored gem hasn't seen any activity in over 7 years, colored2 is a maintained fork, more details: https://libraries.io/rubygems/colored2

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -18,7 +18,7 @@ PATH
       CFPropertyList (~> 2.3.3)
       activesupport (>= 3)
       claide (>= 1.0.1, < 2.0)
-      colored (~> 1.2)
+      colored2 (~> 3.1)
       nanaimo (~> 0.2.3)
 
 GEM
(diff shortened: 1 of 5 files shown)


orta:
It's possible that colored2 doesn't support ruby 2.0?

orta:
Hrm, doesn't imply it - https://github.com/kigster/colored2/blob/master/colored2.gemspec#L20

andrew:
Ah I didn't realise you needed support for both, will investigate, I'm sure we can add that back in

andrew:
Oh you already did it, nice!

orta:
:D

I had looked at doing this for danger before, but needed a push

andrew:
Updated to require version `~> 3.1` 🍰

DanToml:
This is great, thanks!

orta:
Other places we need this: cork, cocoapods (danger)

PR #465, opened 2017-02-03
Add additional file types for extensions

keith:
This adds a few Xcode file types for some extensions that were
previously not included. Before this change if you had a file reference
in a generated project with one of these extensions Xcode would insert
it into the project file after opening it the first time.


segiddins:
👍 with a CHANGELOG entry

keith:
Sorry about that! Added!

PR #466, opened 2017-02-10
Change return value of adding system framework

keith:
Previously it looked like this function was intending to return a file
reference of the newly created framework, but instead it was returning
the original array of strings. This change returns the array of file
references so consumers can manipulate them as needed.

### DIFF
--- a/lib/xcodeproj/project/object/native_target.rb
+++ b/lib/xcodeproj/project/object/native_target.rb
@@ -305,10 +305,11 @@ module Xcodeproj
         #         this reason the new Xcodeproj behaviour is to add the
         #         frameworks in a subgroup according to the platform.
         #
-        # @return [void]
+        # @return [Array<PBXFileReference>] An array of the newly created file
+        #         references.
         #
         def add_system_framework(names)
-          Array(names).each do |name|
+          Array(names).map do |name|
             case platform_name
             when :ios
               group = project.frameworks_group['iOS'] || project.frameworks_group.new_group('iOS')
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 ##### Enhancements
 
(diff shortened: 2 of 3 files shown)


keith:
It doesn't look like CP uses the return value from this function anywhere, so no other changes are required. But this is a breaking change so I'm not sure if we want to do this or not. If we do I'll add a CHANGELOG entry.

keith:
If this behavior is desired I'll have to update the docstrings as well

supermarin:
@keith why is this a breaking change?

keith:
I guess it's not, I thought that it was intentionally returning the input array of strings, but that's actually just a side effect from ruby I guess, according to the docstring.

keith:
I got tricked by the `ref` at the bottom of the each into thinking it was trying to return that.

segiddins:
👍🏻 with a changelog entry

lib/xcodeproj/project/object/native_target.rb
```diff
@@ -308,7 +308,7 @@ def new_shell_script_build_phase(name = nil)
         # @return [void]
```
alloy: The documented return type was `void`, so nobody _should_ have been expecting a stable return value before. So this is not a breaking change 👍

alloy:
It looks like the `ref` at the bottom was left-over from previous code that did indeed return the reference https://github.com/CocoaPods/Xcodeproj/commit/092a12526563550eb70c5c479bc2cea37b43a030#diff-adc2f4667b1ea4b75e763b2aa6cd2b30L146.

supermarin:
@keith please update documentation return type

lib/xcodeproj/project/object/native_target.rb
keith: 🎉

keith:
Updated. Would love a second look!

Issue #180, opened 2014-08-23
[NativeTarget] Add method to read the build settings taking into account any xcconfig


fabiopelosin:
This should be adopted by CocoaPods for the inspection of the targets that will be integrated.

Related: https://github.com/CocoaPods/CocoaPods/issues/2409

segiddins:
@samdmarshall I can't remember, have you or @alloy built a crazy config parser yet?

samdmarshall:
Yes, but it was in python, so i don't think it would be of tremendous value here.

PR #476, opened 2017-04-01
Make build settings parsing optionally take into account any associated xcconfig files

Ruenzuo:
There was a TODO in `AbstractTarget.resolved_build_setting` to implement the necessary logic to take in consideration any associated xcconfig files to targets or projects. While implementing this I realised that adding support for `$(inherited)` is basically a must for this addition to be really useful. I'm not sure if the inheritance rules are documented anywhere (besides the [awesome docs](https://pewpewthespells.com/blog/xcconfig_guide.html#BuildSettingInheritance) provided by @samdmarshall), so this logic is based on what I could see in Xcode 8.3 (check [sample project](https://github.com/Ruenzuo/xcodeproj_parse_xcconfig_settings/tree/master) for reference).

This PR is related to https://github.com/CocoaPods/Xcodeproj/issues/180


CHANGELOG.md
```diff
@@ -4,7 +4,9 @@
 
 ##### Enhancements
 
-* None.  
+* Make build settings parsing optionally take into account any associated xcconfig files.  
+  [Renzo Crisóstomo](https://github.com/ruenzuo)
+  [#476](https://github.com/CocoaPods/Xcodeproj/pull/476)
```
segiddins: This should link to #180

lib/xcodeproj/project/object/build_configuration.rb
```diff
@@ -68,10 +68,34 @@ def type
...
+        # file present and resolving inheritance between them
+        #
+        # @param [String] key
+        #        the key of the build setting.
+        #
+        # @return [String] The value of the build setting
+        #
+        def resolve_build_setting(key)
+          return build_settings[key] if base_configuration_reference.nil?
+          @config ||= Xcodeproj::Config.new(File.new(base_configuration_reference.real_path)).to_hash
```
segiddins: Maybe extract this into its own method?

lib/xcodeproj/project/object/build_configuration.rb
```diff
@@ -68,10 +68,34 @@ def type
...
+          return @config[key] if build_settings[key].nil?
+          return expand_build_setting(build_settings[key], @config[key]) if build_settings[key].include? '$(inherited)'
```
segiddins: I believe `${inherited}` also works

segiddins:
This also needs tests for OTHER_LDFLAGS, which I believe we have separate code paths for, as well as xcconfigs that use conditional settings

lib/xcodeproj/project/object/build_configuration.rb
Ruenzuo: Didn't know that, I'll give it a try.

lib/xcodeproj/project/object/build_configuration.rb
Ruenzuo: Sure.

Ruenzuo:
Completely forgot that conditional settings existed, guess that shows how much I use those 😅. I'll do the required changed to support that, conditional assignment based on configuration sounds rather important for correctness here.

Ruenzuo:
@segiddins I think I implemented all requested changes, with the exception of supporting xcconfig files that use conditional settings. Apparently only `sdk` and `arch` conditions are widely used, but `resolved_build_setting` method returns values [grouped by configuration](https://github.com/CocoaPods/Xcodeproj/blob/master/lib/xcodeproj/project/object/native_target.rb#L44-L45) so I don't know how to implement this without changing the API, which I'm not sure if it's something desired.

Most of the new behaviour is opt-in, with the exception of resolving inherited values. Should I also hide this behind another parameter / method?

lib/xcodeproj/constants.rb
```diff
@@ -382,5 +382,12 @@ module Constants
     # @return [Array] The extensions which are associated with header files.
     #
     HEADER_FILES_EXTENSIONS = %w(.h .hh .hpp .ipp .tpp .hxx .def).freeze
+
+    # @return [Array] The keywords Xcode use to identify a build setting can
+    #                 inherit values from a previous precedence level
+    INHERITED_KEYWORDS = %w(
```
segiddins: @samdmarshall do you know if this encompasses all of the ways of spelling `inherited` in build settings?

samdmarshall: yeah i believe so

orta:
👍

dnkoutso:
@DanToml @orta any objections on merging this? Seems 👍

orta:
Yeah - go for it

PR #479, opened 2017-04-05
Added "inherited" to SWIFT_ACTIVE_COMPILATION_CONDITIONS in Swift debug.

### LINKED ISSUE #478: SWIFT_ACTIVE_COMPILATION_CONDITIONS are ignored
In version 1.4.3 are `SWIFT_ACTIVE_COMPILATION_CONDITIONS` ignored, which causes various CocoaPods to fail after an `pod update`.

Example: https://github.com/delba/Permission

louisdh:
By adding `$(inherited)`, this should no longer override all `SWIFT_ACTIVE_COMPILATION_CONDITIONS` when building in Swift debug. 
This fixes #478.

### DIFF
--- a/lib/xcodeproj/constants.rb
+++ b/lib/xcodeproj/constants.rb
@@ -200,7 +200,7 @@ module Xcodeproj
       }.freeze,
       [:debug, :swift] => {
         'SWIFT_OPTIMIZATION_LEVEL'            => '-Onone',
-        'SWIFT_ACTIVE_COMPILATION_CONDITIONS' => 'DEBUG',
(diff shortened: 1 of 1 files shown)


louisdh:
Specs fail because the Xcode templates by Apple don't include `$(inherited)` for `SWIFT_ACTIVE_COMPILATION_CONDITIONS`. But **we** really do need it, not sure what the best solution is.

The build settings that are compared for the specs are generated by `rake common_build_settings:rebuild`, so we probably shouldn't edit those.

segiddins:
We purposefully match the Xcode defaults. If you're doing this to address an issue in cocoapods, I recommend making the fix there instead of here.

PR #467, opened 2017-02-19
Adds ui_test_bundle extension

1nput0utput:
This is needed to put the right extension in xcake

PR #481, opened 2017-04-07
Add CHANGELOG entry for #467

dnkoutso:


CHANGELOG.md
```diff
@@ -4,6 +4,10 @@
 
 ##### Enhancements
 
+* Adds ui_test_bundle extension.
```
segiddins: this should be more specific, saying that we treat `ui_test_bundle` products as `xctest` bundles

dnkoutso: Only went by the commit message but will change.

orta:
👍

PR #489, opened 2017-05-31
[BuildConfiguration] Take into account project settings when resolving

segiddins:


segiddins:
Technically, for this to be fully comprehensive, it would need to fall back on the platform defaults, but that seems like a lot of work to get right, not to mention being performance-intensive

PR #454, opened 2016-11-23
Add `command_line_arguments` to `TestAction`

brentleyjones:
The documentation for `CommandLineArguments` says
> It can either appear on a LaunchAction or TestAction scheme group.

yet, it wasn't included on `TestAction`. This just adds the same code that is defined in `LaunchAction`.

PR #492, opened 2017-06-02
Add `command_line_arguments` to `TestAction`

DanToml:
Carry forward of \#454


dnkoutso:
I believe only a `CHANGELOG` entry is missing

DanToml:
@dnkoutso yep, was just using CI rather than running tests locally :joy:

PR #501, opened 2017-07-14
Resolve variable substitution for xcconfig declared build settings

Ruenzuo:
Currently, given an xcconfig like:

```
PRODUCT_BUNDLE_IDENTIFIER_Release = com.cocoapods.app
PRODUCT_BUNDLE_IDENTIFIER_Debug = com.cocoapods.app.dev
PRODUCT_BUNDLE_IDENTIFIER = $(PRODUCT_BUNDLE_IDENTIFIER_$(CONFIGURATION))
```

Using `XCBuildConfiguration#resolve_build_setting` for `PRODUCT_BUNDLE_IDENTIFIER ` would return `$(PRODUCT_BUNDLE_IDENTIFIER_$(CONFIGURATION))`. With this change, it returns the proper value `com.cocoapods.app` or `com.cocoapods.app.dev` depending of the configuration.


lib/xcodeproj/project/object/build_configuration.rb
```diff
@@ -103,6 +104,18 @@ def expand_build_setting(build_setting_value, config_value)
           build_setting_value.map { |value| Constants::INHERITED_KEYWORDS.include?(value) ? inherited : value }.flatten
         end
 
+        def resolve_variable_substitution(config_setting)
+          expression = /\$\((.*)\)/
+          match_data = config_setting.match(expression)
+          if match_data.nil?
+            return name if config_setting.eql?('CONFIGURATION')
```
Ruenzuo: Not sure if this is an exhaustive way to resolve the value, would love feedback on this.

CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

```markdown
* Resolve variable substitution for xcconfig declared build settings  
  [Ruenzuo](https://github.com/Ruenzuo)
```

</p>

lib/xcodeproj/project/object/build_configuration.rb
```diff
@@ -82,6 +82,7 @@ def type
         def resolve_build_setting(key)
           setting = build_settings[key]
           config_setting = base_configuration_reference && config[key]
+          config_setting = resolve_variable_substitution(config_setting) if !config_setting.nil? && config_setting.is_a?(String)
```
segiddins: why do we need to check `is_a?(String)`?

Ruenzuo: xcconfig file parsing goes through `XCBuildConfiguration#normalize_array_settings` and we get array values from there. I'm completely ignoring those, as I believe it's mostly used for `OTHER_LDFLAGS`, but I don't fully understand why those have to be arrays, I assume this is expected in other parts of system.

Ruenzuo:
@segiddins not sure if I get the question correctly. Previously `$(CONFIGURATION)` was not expanded at all (at least not by this method), now it is correctly expanded; so yes, it does.

segiddins:
But does it effect any other setting names?

Ruenzuo:
@segiddins Yes. It basically should capture anything between `{()}` and try to resolve the value of that, respecting the level rules in `XCBuildConfiguration#resolve_build_setting`, if no value is found it will just fallback to the config name.

Sorry about the rebase, [someone reported](https://github.com/fastlane/fastlane/pull/9760#issuecomment-315705082) a problem with my implementation. I'll add more tests to try to cover all the different scenarios.

segiddins:
This looks reasonable to me. 👍 
I think I'd like @samdmarshall to chime in if there are known corner cases we should test against for variable expansion, and @DanToml to review the changes.
Thanks!

samdmarshall:
The major thing to watch out for is that the variables you are attempting to expand are actually valid as [xcconfig identifiers](https://pewpewthespells.com/blog/xcconfig_guide.html#Syntax). Otherwise you are going to run into problems with invalid xcconfig files and people wondering why their unicode configuration names aren't working as expected.

Ruenzuo:
Some changes I did after the last review:

* Applied @samdmarshall suggestion, the regex now looks only for valid xcconfig identifiers.
* Made the regex lazy so cases like `USER_DEFINED)_$(CONFIGURATION)` work now.
* Added more tests.

Thanks to the feedback on https://github.com/fastlane/fastlane/pull/9760#issuecomment-316013952 I realised that there's a problem with this approach:

If a build setting defined at a project level uses variable substitution referencing values at a target level the method won't work. I'm still looking into this. Ideally I'd like to jump to the target level from [this point](https://github.com/Ruenzuo/Xcodeproj/blob/master/lib/xcodeproj/project/object/build_configuration.rb#L113) (if `self == project.build_configuration_list[name]`), but not sure if this is possible without passing a reference of the target from the root of the recursion which is something I want to avoid.

Ruenzuo:
@segiddins @DanToml anything else I can do here to move forward this?

lib/xcodeproj/project/object/build_configuration.rb
```diff
@@ -77,15 +77,20 @@ def type
...
         #
+        # @param [PBXNativeTarget] root_target
+        #        use this to resolve complete recursion between project and targets
+        #
         # @return [String] The value of the build setting
         #
-        def resolve_build_setting(key)
+        def resolve_build_setting(key, root_target = nil)
           setting = build_settings[key]
+          setting = resolve_variable_substitution(setting, root_target) if !setting.nil? && setting.is_a?(String)
```
segiddins: checking both `nil?` and `is_a?` is redundant

lib/xcodeproj/project/object/build_configuration.rb
```diff
@@ -77,15 +77,20 @@ def type
...
           config_setting = base_configuration_reference && config[key]
+          config_setting = resolve_variable_substitution(config_setting, root_target) if !config_setting.nil? && config_setting.is_a?(String)
```
segiddins: checking both nil? and is_a? is redundant

lib/xcodeproj/project/object/build_configuration.rb
```diff
@@ -103,6 +108,20 @@ def expand_build_setting(build_setting_value, config_value)
           build_setting_value.map { |value| Constants::INHERITED_KEYWORDS.include?(value) ? inherited : value }.flatten
         end
 
+        def resolve_variable_substitution(config_setting, root_target)
+          expression = /\$[{(]([^inherited][$(){}_a-zA-Z0-9]*?)[})]/
```
segiddins: extract this into a constant and please comment the regexp

lib/xcodeproj/project/object/build_configuration.rb
Ruenzuo: Makes totally sense!

lib/xcodeproj/project/object/build_configuration.rb
Ruenzuo: Makes totally sense!

lib/xcodeproj/project/object/build_configuration.rb
Ruenzuo: Done.

segiddins:
I'll let @DanToml merge this one

lib/xcodeproj/project/object/build_configuration.rb
```diff
@@ -96,13 +101,42 @@ def resolve_build_setting(key)
 
         private
 
+        CAPTURE_VARIABLE_IN_BUILD_CONFIG = /
```
DanToml: thanks for documenting the regex 👍

litalhassine:
Any plans for a new release with this commit?

dnkoutso:
@litalhassine you can use http://bundler.io/ and point to a specific SHA or branch to use this.

Ruenzuo:
@dnkoutso it would be really nice to have this as a patch release (probably nobody is using that API anyway) so we could submit a fix for the issue described here: https://github.com/fastlane/fastlane/pull/9460#issuecomment-312931119.

casz:
Thanks this resolved one of the issue I had in my 
https://github.com/casz/xcodeproj-sample-env
the case [`The subtitues variables that are resolved by xcode build settings`](https://github.com/casz/xcodeproj-sample-env/blob/4a7020d134133a3c96d124bc98c1ad09d8f19125/env_spec.rb#L19-L31)
Does not work with xcodeproj 1.5.1

However it works in 1.5.2, great work 🚀

epatel:
This don't seem to work for me. I am using fastlane and it always fails on export now, did work earlier though (Xcode8? Xcode9 now).

I have `$(TEST_GROUP)` that can be defined as a couple of test groups and then a number of other variables, in a `BUNDLE_ID` that is a concatenation like `$($(TEST_GROUP)_BUNDLE_ID)` which then gets i.e. from `EXTERNAL_BUNDLE_ID = com.myapp.external` if `TEST_GROUP` is set to `EXTERNAL`.

I see `$($(TEST_GROUP)_BUNDLE_ID)` becomes `TEST_GROUP_BUILD_BUNDLE_ID`

This worked for a while and now it does not...
http://dyna.mo/xcconfig/

Ruenzuo:
@epatel I'm happy to have a look if you could open an issue with a sample project to reproduce this.

PR #514, opened 2017-10-26
Support *.cc file as a header

sharkpc0813:
Make PR for [issue](https://github.com/CocoaPods/CocoaPods/issues/6833) again because of some conflicts..

`*.cc` file would be included exceptionally.

https://google.github.io/styleguide/cppguide.html#Self_contained_Headers


segiddins:
The above link doesn't seem to imply that `.cc` files are to be interpreted as headers

sharkpc0813:
The definitions for template and inline functions can be placed in `cc`file and pod couldn't read this.

segiddins:
> As an exception, a template that is explicitly instantiated for all relevant sets of template arguments, or that is a private implementation detail of a class, is allowed to be defined in the one and only .cc file that instantiates the template.

This seems to say that the _only_ time it's ok to put a template implementation in a `.cc` file is when that file is the only one to instantiate the template

sharkpc0813:
I think you may concern about `.cc` file may not contain above of this. 
But the file which contain templates and inline functions is used like header file. This would be included like  `#include<things.cc>`.

CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Please rebase to get rid of the merge commits in this PR</td>
    </tr>
  </tbody>
</table>

</p>

PR #525, opened 2017-11-21
Maintain main_group annotations on save

### LINKED ISSUE #524: main_group annotations are being removed on save
I am looking in to using this project for some simple CLI tooling on xcode projects.
Some projects have an annotation on their mainGroup e.g. some of WebKit's xcode projects.
https://raw.githubusercontent.com/WebKit/webkit/master/Source/WebCore/WebCore.xcodeproj/project.pbxproj
I can't link directly, but here's the lines with the annotations:
`mainGroup = 0867D691FE84028FC02AAC07 /* WebKit */;`
`0867D691FE84028FC02AAC07 /* WebKit */ = {`
These annotations are currently being removed on a save.

I'm not sure if this is some legacy xcode feature but it would be nice to maintain these annotations on a project save.

c29reid:
As described in #524 some projects have an annotation on their main_group which are being removed when the project is loaded and saved by the Xcodeproj library.

This fixes #524

I am new to this project so let me know if there's anything else I should be doing.
Thanks

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@
   [izaakschroeder](https://github.com/izaakschroeder)
   [#513](https://github.com/CocoaPods/Xcodeproj/issues/513)
 
+* Don't remove the main_group annotations on a save
+  [c29reid](https://github.com/c29reid)
+  [#524](https://github.com/CocoaPods/Xcodeproj/issues/524)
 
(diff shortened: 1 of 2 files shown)


c29reid:
I didn't realize it now adds a default annotation, I can look in to that

dnkoutso:
PR outdated.

Issue #531, opened 2017-12-18
undefined method `test_target_type?'


yadirhb:
Hi there I'm trying to recreate the user schemes for a project using the demo specification as follows: 
```
require 'xcodeproj'
xcproj = Xcodeproj::Project.open('MyApp.xcodeproj')
xcproj.recreate_user_schemes
xcproj.save
```
but I am getting this error: 

```
 ./fix_schemes.rb
/usr/local/lib/ruby/gems/2.4.0/gems/xcodeproj-1.5.4/lib/xcodeproj/project.rb:812:in `block in recreate_user_schemes': undefined method `test_target_type?' for #<Xcodeproj::Project::Object::PBXAggregateTarget:0x00007fb72fcdd168> (NoMethodError)
        from /usr/local/lib/ruby/gems/2.4.0/gems/xcodeproj-1.5.4/lib/xcodeproj/project.rb:809:in `each'
        from /usr/local/lib/ruby/gems/2.4.0/gems/xcodeproj-1.5.4/lib/xcodeproj/project.rb:809:in `recreate_user_schemes'
        from ./fix_schemes.rb:4:in `<main>'
```
I've already check inside the native_target definition and exists such method but at runtime I don't know why it is failing. I did not found any related explanation online that's why I'm reporting this issue in here. Hope can fix or explain this.

yadirhb:
I've already checked and this fails in my project only if I add a target of type "External Build System", otherwise generates the schemes nicely. 

![image](https://user-images.githubusercontent.com/6015754/34123854-b21e9c4c-e40f-11e7-836f-c06a53cb0618.png)

PR #532, opened 2017-12-18
Fixing the issue #531

yadirhb:
Originally the code was assuming that every target had the method "test_target_type" but looking into the targets definition on file (xcodeproj/project/object/native_target.rb) I've notice that there are three target types:
1-PBXNativeTarget
2-PBXAggregateTarget
3-PBXLegacyTarget 

The only target type which defines the method "test_target_type" is PBXNativeTarget, that is why the change proposal adds the type verification check before actually call the referred method.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@
 
 ##### Bug Fixes
 
-* Fixing the method recreate_user_scheme for targets other than type PBXNativeTarget.
+* Fixing the method recreate_user_scheme for targets other than type PBXNativeTarget.  
   [Yadir Hernandez](https://github.com/yadirhb)
   [#531](https://github.com/CocoaPods/CocoaPods/issues/531)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,10 @@
 
 ##### Bug Fixes
 
+* Fixing the method recreate_user_scheme for targets other than type PBXNativeTarget.
+  [Yadir Hernandez](https://github.com/yadirhb)
+  [#531](https://github.com/CocoaPods/CocoaPods/issues/531)
+
 * Verify container portal when checking dependency target membership.  
   [izaakschroeder](https://github.com/izaakschroeder)
   [#513](https://github.com/CocoaPods/Xcodeproj/issues/513)
--- a/lib/xcodeproj/project.rb
+++ b/lib/xcodeproj/project.rb
@@ -809,7 +809,7 @@ module Xcodeproj
       targets.each do |target|
         scheme = XCScheme.new
         scheme.add_build_target(target)
-        scheme.add_test_target(target) if target.test_target_type?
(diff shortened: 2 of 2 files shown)


CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

```markdown
* Fixing the issue #531  
  [yadirhb](https://github.com/yadirhb)
```

</p>

dnkoutso:
Seems good. Can you please add a changelog entry and change your commit message to be more descriptive? Like "do not add test target references to aggregate targets"

yadirhb:
This was a bug fix, so now on if targets are other than PBXNativeTarget the method recreate_user_scheme will work as well. I think this is the description we’re looking for.

Enviado desde Correo para Windows 10

De: Dimitris Koutsogiorgas
Enviado: Monday, December 18, 2017 5:56 PM
Para: CocoaPods/Xcodeproj
CC: Yadir Hernandez Batista; Author
Asunto: Re: [CocoaPods/Xcodeproj] Fixing the issue #531 (#532)

Seems good. Can you please add a changelog entry and change your commit message to be more descriptive? Like "do not add test target references to aggregate targets"

lib/xcodeproj/project.rb
```diff
@@ -809,7 +809,7 @@ def recreate_user_schemes(visible = true)
       targets.each do |target|
         scheme = XCScheme.new
         scheme.add_build_target(target)
-        scheme.add_test_target(target) if target.test_target_type?
+        scheme.add_test_target(target) if target.instance_of?(Object::PBXNativeTarget) && target.test_target_type?
```
dnkoutso: we might want to implement this `if target.responds_to?(:test_target_type?) && target.test_target_type?`

Also needs a test

CHANGELOG.md
```diff
@@ -21,6 +21,10 @@
 
 ##### Bug Fixes
 
+* Fixing the method recreate_user_scheme for targets other than type PBXNativeTarget.
```
dnkoutso: please add two trailing spaces at the end of this string.

lib/xcodeproj/project.rb
yadirhb: I just really gave you an idea on what's the problem and a possible solution. I solved this issue on my own, by processing the targets exactly the same way proposed in here. Until a new version be released with this bug fixed. This is my first experience with Ruby ever, so can't tell much more.

Thanks a lot for the support, and for leting me contribute with this project.

yadirhb: Tested responds_to method as follows:

image

dnkoutso: yes sorry please use `respond_to?` instead

yadirhb: Ready

dnkoutso:
LGTM.

PR #533, opened 2017-12-18
Adding changelog for issue 531 resolution

yadirhb:


dnkoutso:
Still wrong. Please update the original PR. Otherwise it will not turn green without changelog entry. You can have multple commits in a single PR.

PR #543, opened 2018-01-30
Create a buildable reference to another project's target

simonseyer:
Since the buildable reference was not aware of its own project, it used the target's project to construct a buildable reference. By default, this behavior still applies. However, now it is possible to pass in a `root_project` which is used to create the reference (`root_project` -> `target.project`).

### DIFF
--- a/lib/xcodeproj/scheme/buildable_reference.rb
+++ b/lib/xcodeproj/scheme/buildable_reference.rb
@@ -110,12 +110,13 @@ module Xcodeproj
       #                  the actual project directory name]"
       #
       def construct_referenced_container_uri(target, root_project = nil)
-        project = root_project || target.project
         target_project = target.project
-        path = if !project.root_object.project_dir_path.to_s.empty?
-                 project.path + project.root_object.project_dir_path
+        root_project ||= target_project
+        root_project_dir_path = root_project.root_object.project_dir_path
+        path = if !root_project_dir_path.to_s.empty?
+                 root_project.path + root_project_dir_path
                else
-                 project.path.dirname
+                 root_project.path.dirname
                end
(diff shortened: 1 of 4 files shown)


lib/xcodeproj/scheme/buildable_reference.rb
```diff
@@ -96,14 +102,23 @@ def construct_buildable_name(build_target)
...
       #                  the actual project directory name]"
       #
-      def construct_referenced_container_uri(target)
-        project = target.project
-        relative_path = project.path.relative_path_from(project.path + project.root_object.project_dir_path).to_s
-        relative_path = project.path.basename if relative_path == '.'
+      def construct_referenced_container_uri(target, root_project = nil)
+        project = root_project || target.project
+        target_project = target.project
+        path = if !project.root_object.project_dir_path.to_s.empty?
```
segiddins: extract `project.root_object.project_dir_path` into a variable?

lib/xcodeproj/scheme/buildable_reference.rb
```diff
@@ -96,14 +102,23 @@ def construct_buildable_name(build_target)
...
       #                  file relative to the project_dir_path, always ending with
       #                  the actual project directory name]"
       #
-      def construct_referenced_container_uri(target)
-        project = target.project
-        relative_path = project.path.relative_path_from(project.path + project.root_object.project_dir_path).to_s
-        relative_path = project.path.basename if relative_path == '.'
+      def construct_referenced_container_uri(target, root_project = nil)
+        project = root_project || target.project
+        target_project = target.project
```
segiddins: move this up a line, and use it when assigning to `project`?

segiddins:
Great, thanks!

👍

PR #547, opened 2018-02-08
[CI] Test against new Ruby 2.5 and patched

nicolasleger:


nicolasleger:
Some other gems or projects my depend on it like CocoaPods

segiddins:
And CocoaPods only tests on 3 ruby versions

PR #548, opened 2018-02-14
[BuildConfiguration] Fix expanding settings when the inherited value …

segiddins:
…is an array but the current value is a string

Closes https://github.com/CocoaPods/CocoaPods/issues/7421.

cc @jeffreyffs


dnkoutso:
Just good old rubocop

Issue #566, opened 2018-04-12
Config parser does not recognize assignments without whitespace


stocc:
In my project, I have several xcconfig files that have assignments without spaces separating the equal sign like
`PRODUCT_NAME=$(CUSTOMER_PRODUCT_NAME)`
Xcode parses these just fine but the regex used here appears to enforce whitespace

https://github.com/CocoaPods/Xcodeproj/blob/aac57c46d85b502be2defec4211c925b196966d1/lib/xcodeproj/config.rb#L24

Simply the `+`should be replaced by `*`

PR #567, opened 2018-04-13
Fixes regex parsing config files

stocc:
See #566


spec/fixtures/differing-whitespace.xcconfig
```diff
@@ -0,0 +1,3 @@
+Key1=Value1
+Key2 = Value2
+Key3  = Value3
```
segiddins: Would you mind adding tests that include `[]` specifiers on the key?

segiddins:
Thanks!

PR #580, opened 2018-06-19
Only require 'digest' once instead of on every call of `uuid_for_path`

amorde:
Although Ruby does cache the results of `require`, this still appears to be accounting for a significant portion of `pod install`. It's called over 10,000 times on a project I am testing.

This reduced the `pod install` time on a test project from ~12s to ~10s.


dnkoutso:
LGTM

PR #581, opened 2018-06-23
Reduce the number of string allocations

amorde:
cc @segiddins

### DIFF
--- a/lib/xcodeproj/project.rb
+++ b/lib/xcodeproj/project.rb
@@ -46,6 +46,10 @@ module Xcodeproj
     #
     attr_reader :path
 
+    # @return [Pathname] the directory of the project
+    #
+    attr_reader :project_dir
+
     # @param  [Pathname, String] path @see path
     #         The path provided will be expanded to an absolute path.
     # @param  [Bool] skip_initialization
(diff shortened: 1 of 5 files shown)


lib/xcodeproj/project/object_attributes.rb
```diff
@@ -354,14 +354,15 @@ def has_one(singular_name, isas)
             # 1.9.2 fix, see https://github.com/CocoaPods/Xcodeproj/issues/40.
             public(attrb.name)
 
+            variable_name = "@#{attrb.name}"
```
amorde: The attribute method definitions were allocating over 360k instances of "@dependencies" and "dependencies", so I thought this was worth it :)

lib/xcodeproj/project.rb
```diff
@@ -64,6 +68,7 @@ class Project
     #
     def initialize(path, skip_initialization = false, object_version = Constants::DEFAULT_OBJECT_VERSION)
       @path = Pathname.new(path).expand_path
+      @project_dir = @path.dirname
```
amorde: Cuts down on some of the allocations of the `Pods/` directory

lib/xcodeproj/project/object_attributes.rb
segiddins: nit: `variable_name = :"@#{attrb.name}"` so it's a symbol instead of a string

amorde: fixed

Issue #618, opened 2018-10-24
Scheme's configure_with_targets setting test targets as build targets in 1.7.0


joshuapoq:
Using `scheme.configure_with_targets(nil, test_target)` causes test targets in the scheme to have archiving enabled.

I believe this was caused by https://github.com/CocoaPods/Xcodeproj/pull/614 but causes test targets that are added to have archiving, profiling and building enabled. See below screenshots.

**Xcodeproj - 1.6.0:**
![xcodeproj-1 6 0](https://user-images.githubusercontent.com/28047117/47423906-35026280-d77e-11e8-9438-3ed771a8a6ce.png)

**Xcodeproj - 1.7.0:**
![xcodeproj-1 7 0](https://user-images.githubusercontent.com/28047117/47423904-35026280-d77e-11e8-8429-093980a95190.png)

PR #622, opened 2018-11-02
[#618] Fix test entries scheme options

dostrander:
@Kapin @segiddins @dnkoutso 

Fixes: https://github.com/CocoaPods/Xcodeproj/issues/618

### DIFF
--- a/lib/xcodeproj/scheme.rb
+++ b/lib/xcodeproj/scheme.rb
@@ -193,9 +193,9 @@ module Xcodeproj
 
       entry.build_for_testing   = true
       entry.build_for_running   = build_for_running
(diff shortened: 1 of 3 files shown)


lib/xcodeproj/scheme.rb
```diff
@@ -78,6 +78,15 @@ def configure_with_targets(runnable_target, test_target, launch_target: false)
       end
     end
 
+    # def configure_with_targets(runnable_target, test_target)
```
dnkoutso: delete instead?

dostrander: Ooops yep

PR #652, opened 2019-01-10
Update default script phase values to match Xcode 10.

dnkoutso:
Xcode 10 (actually Xcode 9 too) never set `showEnvVarsInLog=1` inside a `pbxproj` even though the box is checked.

Additionally Xcode 10 has a different default value that is a comment and transcribed into the `pbxproj` itself under the script phase.

No need for a new test as its already covered.


lib/xcodeproj/project/object/build_phase.rb
```diff
@@ -279,16 +279,14 @@ class PBXShellScriptBuildPhase < AbstractBuildPhase
 
         # @return [String] the actual script to perform.
         #
-        # @note   Defaults to the empty string.
+        # @note   Defaults to a comment string.
         #
-        attribute :shell_script, String, ''
+        attribute :shell_script, String, "# Type a script or drag a script file from your workspace to insert its path.\n"
```
justinseanmartin: I'm not sure that Xcodeproj should care about imitating this behavior of Xcode. I assume this default exists mostly for Xcode UI discoverability/usability, but isn't terribly helpful for this gem. I don't feel strongly about this though.

lib/xcodeproj/project/object/build_phase.rb
```diff
@@ -279,16 +279,14 @@ class PBXShellScriptBuildPhase < AbstractBuildPhase
...
 
         # @return [String] whether or not the ENV variables should be shown in
         #         the build log.
         #
-        # @note   Defaults to true (`1`).
-        #
-        attribute :show_env_vars_in_log, String, '1'
+        attribute :show_env_vars_in_log, String
```
justinseanmartin: Should something change such that setting this to '1' is equivalent to clearing it?

lib/xcodeproj/project/object/build_phase.rb
dnkoutso: Its nuanced in a sense that if xcodeproj creates an empty script phase with defaults then Xcode will change it again therefore making any repo that is tracked by version control to be dirty.

dnkoutso: Although yes not many people in reality create empty script phases through Xcodeproj but its OK to fix this here.

dnkoutso: Yes it does at least with Xcode 10. It also adds it as a comment using `#` and sets the value. Compared to Xcode 9 that just shows the same text without `#` but _does not_ set it as part of `pbxproj`. I tested on both and this is why I changed it here for Xcode 10 and going forward.

lib/xcodeproj/project/object/build_phase.rb
dnkoutso: I am not sure we want to add this logic here so much....I am making the PR in CocoaPods to ensure that this does not happen though

PR #655, opened 2019-01-23
add .rbenv-gemsets to gitignore

asmallteapot:
[rbenv-gemset](https://github.com/jf/rbenv-gemset) is helpful for developing gems while also using the stable version.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 *.bundle
 *.o
 *.rbc
+.rbenv-gemsets
 .rbenv-version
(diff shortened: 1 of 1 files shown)


segiddins:
Tbh that probably belongs more in a user-global gitignore than in every project gitignore

asmallteapot:
@segiddins We already ignore the `.rbenv-version` – seems like we should have both or neither…?

dantoml:
Gonna merge this because we also ignore rbenv-versions - ideally folks would have those in their global gitignore, but I'm ok with making common tools easier.

PR #668, opened 2019-03-20
Set root object compatibility version depending on object version.

dnkoutso:
This is out of sync with the `objectVersion` we set for the generated projects by Xcodeproj and ensures they remain in sync.

### DIFF
--- a/lib/xcodeproj/project.rb
+++ b/lib/xcodeproj/project.rb
@@ -80,6 +80,10 @@ module Xcodeproj
       unless skip_initialization
         initialize_from_scratch
         @object_version = object_version.to_s
+        unless Constants::COMPATIBILITY_VERSION_BY_OBJECT_VERSION.key?(object_version)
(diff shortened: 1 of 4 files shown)


lib/xcodeproj/project.rb
```diff
@@ -80,6 +80,7 @@ def initialize(path, skip_initialization = false, object_version = Constants::DE
       unless skip_initialization
         initialize_from_scratch
         @object_version = object_version.to_s
+        root_object.compatibility_version = Constants::COMPATIBILITY_VERSION_BY_OBJECT_VERSION.fetch(object_version)
```
dnkoutso: we use `fetch` here because if a new object_version is passed we want to ensure that the `COMPATIBILITY_VERSION_BY_OBJECT_VERSION` hash is updated.

segiddins: should explicitly test what happens when the object version is unknown, and ensure that a helpful error message is raised (with instructions to file an issue)

dnkoutso: done

PR #669, opened 2019-03-27
Use modern localization identifier 'en' for the development region

HeEAaD:
Change default development_region from "English" to "en" because "English" is deprecated.

# Issue
From the Xcode 10.2 release notes:
> Xcode now more carefully distinguishes between legacy localization identifiers such as “English” and modern localization identifiers such as “en” and represents them both in both project files and the user interface. (45469882)

https://developer.apple.com/documentation/xcode_release_notes/xcode_10_2_release_notes?language=objc#3136790

This leads to deprecation warnings in Xcode 10.2:
> Migrating the English, deprecated localization to English is recommended for all projects. This will ensure localized resources are placed in “en.lproj” directories instead of deprecated “English.lproj” directories.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@
 * Normalize xcconfig path when generating includes.  
   [bclymer](https://github.com/bclymer)
 
+* Use modern localization identifier 'en' for the development region.
+  [Steffen Matthischke](https://github.com/heeaad)
+
 ## 1.8.1 (2019-02-19)
 
 ##### Enhancements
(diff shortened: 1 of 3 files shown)


amorde:
LGTM but would like 1 more review.

I've experienced this in my own projects as well

amorde:
Thanks @HeEAaD!

CHANGELOG.md
```diff
@@ -15,6 +15,9 @@
 * Normalize xcconfig path when generating includes.  
   [bclymer](https://github.com/bclymer)
 
+* Use modern localization identifier 'en' for the development region.
```
dnkoutso: change log entry was not rebased to latest master it now appears as if it was in 1.8.2 will fix.

PR #677, opened 2019-04-25
Sorting like Xcode. Fixes #615.

### LINKED ISSUE #615: Sort command does not use the same sorting scheme as Xcode
Xcode supports sorting group by right clicking on the group and choosing "Sort by Name".
If you have a group with two files  - `A10.h` and `A2.h`, for example, Xcode will put `A2.h` first.
`xcodeproj sort` command will do the simple alphabetic sort and put `A10.h` first.

The scheme used by Xcode is weird (try sorting A1 and A01, or even better A1B01, and A01B1). If someone accidentally knows what's the exact sorting scheme it will be great, but fixing the most simple case is sufficient.

Coeur:
I'm not fully familiar with Ruby (apart from a Podfile/Podspec), so advices are welcome.


Coeur:
OK, I saw the RuboCop errors: I'll fix those soon.

Coeur:
Hum, actually, my algorithm doesn't match Xcode behavior fully. Let me see if I can improve it.

lib/xcodeproj/project/object/group.rb
```diff
@@ -437,10 +438,9 @@ def sort(options = nil)
                 next groups_position == :above ? 1 : -1
               end
             end
-
-            result = File.basename(x.display_name.downcase, '.*') <=> File.basename(y.display_name.downcase, '.*')
+            result = XcodeSortString.new(File.basename(x.display_name, '.*')) <=> XcodeSortString.new(File.basename(y.display_name, '.*'))
```
segiddins: should `XcodeSortString` internally handle sorting lexicographically by basename and then extname ?

spec/project/object/helpers/sort_helper_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+    before do
+      @helper = XcodeSortString
+    end
+
+    #-------------------------------------------------------------------------#
+
+    describe 'In general' do
+      it 'sorts names like Xcode' do
+        unsorted_names = [
+          # spaces comparison (test disabled: not properly supported)
```
segiddins: should this be handled?

lib/xcodeproj/project/object/helpers/sort_helper.rb
```diff
@@ -0,0 +1,34 @@
...
+    include Comparable
+    attr_reader :str, :scrubbed, :ints_and_strings, :i_s_pattern
+
+    def initialize(str)
+      @str = str
+      @scrubbed = str.downcase
+      # `Integer`+`rescue`: credit to https://stackoverflow.com/a/39025521
+      @ints_and_strings = @scrubbed.scan(/\d+|\D+/).map { |s| begin Integer(s, 10); rescue ArgumentError; s end }
+      # comparing patterns: credit to https://rosettacode.org/wiki/Natural_sorting#Ruby
+      @i_s_pattern = @ints_and_strings.map { |el| el.is_a?(Integer) ? :i : :s }.join
```
segiddins: could you please give this ivar a more descriptive name?

lib/xcodeproj/project/object/helpers/sort_helper.rb
```diff
@@ -0,0 +1,34 @@
...
+
+    def <=>(other)
+      if i_s_pattern.start_with?(other.i_s_pattern) || other.i_s_pattern.start_with?(i_s_pattern)
+        compare = ints_and_strings <=> other.ints_and_strings
+        if compare != 0
+          # we sort naturally
+          compare
+        else
+          # equality, we sort reverse raw
+          -(str <=> other.str)
```
segiddins: to sort in reverse order, you can do `other.str <=> str` with no need for negation

lib/xcodeproj/project/object/helpers/sort_helper.rb
```diff
@@ -0,0 +1,34 @@
...
+  #   arrayOfFilenames.sort_by { |s| XcodeSortString.new(s) }
+  class XcodeSortString
+    include Comparable
+    attr_reader :str, :scrubbed, :ints_and_strings, :i_s_pattern
+
+    def initialize(str)
+      @str = str
+      @scrubbed = str.downcase
+      # `Integer`+`rescue`: credit to https://stackoverflow.com/a/39025521
+      @ints_and_strings = @scrubbed.scan(/\d+|\D+/).map { |s| begin Integer(s, 10); rescue ArgumentError; s end }
```
segiddins: Can we do this without having to swallow exceptions?

lib/xcodeproj/project/object/group.rb
Coeur: No, I think this situation is just to handle identical filenames in different folders: let's keep the fallback independent from the sort helper..

lib/xcodeproj/project/object/helpers/sort_helper.rb
Coeur: it's the "integers and strings pattern". I'll try to rename it.

lib/xcodeproj/project/object/helpers/sort_helper.rb
Coeur: Yes, absolutely, by adding a regex match: `{ |s| \d\ === s ? Integer(s, 10) : s }`.
I do not know which one is the most performant, but I thought an exception may be faster than a regex match.
I'll change it.

Coeur:
> What are the performance implications of this change?

The filenames are split on their digits part, then the sort is made on those collections: the cost is probably quite low. We may do a performance test, but project filenames are generally short (often less than 50 chars) and don't often have more than 1 digit part. So I bet it would be at worst sorting twice slower than before.

Coeur:
@segiddins I've now solved all my known test cases.

lib/xcodeproj/project/object/group.rb
Coeur: I've tested this situation with Xcode, and Xcode does not sort at all when filenames are identical: it will simply keep the existing order. That supports that we shouldn't include this behavior in `XcodeSortString` itself, and we should simply keep it externally like that.

Coeur:
@dnkoutso Would you like to review this PR as well?

Coeur:
OK, I figured out something worth mentioning:
**This PR solves the "integer" natural sorting, but it does not solve the "unicode" sorting.**

Solving the unicode sort would most likely require `activesupport` dependency, which was removed two years ago by @segiddins according to the changelog.
So I'd like to know if this PR, as it is currently, is worth merging for the its partial improvement on sorting, or if I need to re-introduce activesupport in Xcodeproj in order to be able to use [`transliterate`](https://github.com/rails/rails/blob/master/activesupport/lib/active_support/inflector/transliterate.rb) or [`mb_chars`](https://github.com/rails/rails/blob/6-0-stable/activesupport/lib/active_support/core_ext/string/multibyte.rb)`.normalize(:kd)` and attempt some Unicode Collation Algorithm sorting?

dnkoutso:
Mostly concerned with performance implications here. Large projects can take a big hit. Can you simulate a scenario with many many files and benchmark it?

Coeur:
@dnkoutso I was busy solving some edge cases, and I figured out that Xcode sorting is broken for Unicode: http://www.openradar.me/radar?id=5012044621283328

I also figured out, from https://travis-ci.org/CocoaPods/Xcodeproj/builds/533316561 that ruby 2.4 and newer will sort Unicode strings differently than in ruby 2.3 and older.

Lastly, while macOS 10.14.5 is bundled with **ruby 2.3.7p456**, the upcoming macOS 10.15 will be bundled with **ruby 2.6.0** according to Google: https://www.google.com/search?q=ruby+macOS+%2210.15%22.

All in all, we should wait for macOS 10.15 beta in a couple of weeks and see if Xcode sorting is affected by the newer version of ruby.

As for performances, I suggest generating 10,000 UDID (easy mix of letters and digits) and measuring sorting that data with a plain `downcase` compare and with our `XcodeSortString` helper. I haven't done the benchmark yet.

dnkoutso:
> As for performances, I suggest generating 10,000 UDID (easy mix of letters and digits) and measuring sorting that data with a plain downcase compare and with our XcodeSortString helper. I haven't done the benchmark yet.

sg!

dnkoutso:
any updates here? Looking to make a release for Xcode 11 this week.

Coeur:
@dnkoutso I'll try to build a performance comparison in the next couple of days.

Coeur:
@dnkoutso :
```ruby
require 'benchmark'
require 'securerandom'

array = (1..10000).map { SecureRandom.uuid }
array1 = array.dup
array2 = array.dup

Benchmark.bmbm do |x|
    x.report("downcase") { array1.sort! do |x, y|; x.downcase <=> y.downcase; end }
    x.report("XcodeSortString") { array2.sort! do |x, y|; Xcodeproj::XcodeSortString.new(x) <=> Xcodeproj::XcodeSortString.new(y); end }
end
```

output:
```
                      user     system      total        real
downcase          0.010000   0.000000   0.010000 (  0.011024)
XcodeSortString  14.490000   0.050000  14.540000 ( 14.556600)
```

lol ... I guess we need to give the choice between a fast sort and an accurate sort.

dnkoutso:
If we can maintain the performance or make it negligible we can land this!

Coeur:
```
Offenses:
lib/xcodeproj/project/object/helpers/sort_helper.rb:8:5: C: Replace class var @@has_unicode_normalize with a class instance var.
    @@has_unicode_normalize = Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('2.2.0')
    ^^^^^^^^^^^^^^^^^^^^^^^
112 files inspected, 1 offense detected
RuboCop failed!
```
Hum, no, I don't want an instance var, I want a static evaluation that is done only once.

@dnkoutso do you know how to workaround RuboCop here?

lib/xcodeproj/project/object/group.rb
Coeur: @segiddins actually, I misunderstood `extname` for "extended name" (aka fullpath) when actually it's a function to get the "file extension". So yes, you're right, there is something to fix here. I'm looking at it.

Coeur:
I just found issues in sorting symbols... so it's not ready to merge.

PR #720, opened 2019-10-17
Fix config-dump array serialization

segiddins:


dnkoutso:
If you'd like this for 1.13.1 please point to `1-13-stable` branch.

segiddins:
It's not fixing a regression, so it should go to master

Issue #744, opened 2020-03-25
Add support for Xcode 11.4 objectVersion/compatibilityVersion


madsolar8582:
Xcode 11.4 adds a new compatibility setting for the project:

```
objectVersion = 53;
compatibilityVersion = "Xcode 11.4";
```

Please add this so that the runtime error is not thrown when the Xcodeproj gem attempts to parse the project file:
```
/Library/Ruby/Gems/2.6.0/gems/xcodeproj-1.15.0/lib/xcodeproj/project.rb:227:in `initialize_from_file': [Xcodeproj] Unknown object version. (RuntimeError)
```

dnkoutso:
Yeap!

PR #687, opened 2019-06-04
Support for Xcode 11 attributes and objects.

dnkoutso:
Closes https://github.com/CocoaPods/CocoaPods/issues/8871
Closes https://github.com/CocoaPods/Xcodeproj/issues/688

PR #746, opened 2020-03-26
Add Xcode 11.4 object version.

dnkoutso:
closes https://github.com/CocoaPods/Xcodeproj/issues/744

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 ##### Enhancements
 
+* Add Xcode 11.4 object version.  
+  [Dimitris Koutsogiorgas](https://github.com/dnkoutso)
+  [#744](https://github.com/CocoaPods/Xcodeproj/issues/744)
+
 * Add new APIs to set testables or entries in schemes.  
   [Dimitris Koutsogiorgas](https://github.com/dnkoutso)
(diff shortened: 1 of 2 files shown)


amorde:
What changed in the project format? Fine merging this but the version wouldn't bump without something changing right?

dnkoutso:
Who knows? Apple bumps that number when they do...it seems from the issue report that Xcode 11.4 has a new version of 53.

amorde:
I was under the impression it only gets bumped if something actually changes (ex. new attribute gets added)

this same topic came up last time https://github.com/CocoaPods/Xcodeproj/pull/687#discussion_r290145951

dnkoutso:
Inteeeeresting....you are correct. I guess this doesnt matter since after we added this we are now more strict and crash for an unknown object version. This PR would fix the crash and we should get warnings for new attributes?

amorde:
Don't we also crash on unknown attributes?

dnkoutso:
No we warn for these.

dnkoutso:
It actually says to file an issue in the warning.

amorde:
As I see https://github.com/CocoaPods/Xcodeproj/blob/80a2cdabd49eaaeff177a132c97b393f7337fbc9/spec/project/object_spec.rb#L165

Would still prefer to actually add support for whatever changed but understand the need to fix crashes sooner rather than later

PR #772, opened 2020-07-07
Override Xcode 12 default for erroring on quoted imports in umbrellas

paulb777:
Fix https://github.com/CocoaPods/CocoaPods/issues/9902


dnkoutso:
@paulb777 any reason not to do that in CocoaPods xcconfig settings instead?

paulb777:
@dnkoutso No. I just didn't know it was possible there and I didn't find the right spot.

dnkoutso:
I think you can do it in https://github.com/CocoaPods/CocoaPods/blob/master/lib/cocoapods/target/build_settings.rb

paulb777:
SG - I'll redo it there.

PR #796, opened 2020-10-15
Add support for pre/post-actions in scheme actions

### LINKED ISSUE #401: Creating pre- and post-build scheme actions
Add support for creating pre-build and post-build scheme actions as part of the scheme support in Xcodeproj.

The caveat here is that @segiddins is not allowed to work on this feature. 🙂

thiagohmcruz:
Resolves #401 

Add support for pre/post-actions in the `Build`, `Test`,  `Profile` and `Archive` scheme actions. Note that the `Analyze` scheme action in Xcode does not support pre/post-actions so it has to be explicitly disabled in `AbstractSchemeAction`.


CocoaPodsBarista:
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true" data-kind="Warning">
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

```markdown
* Creates pre/post-actions for scheme actions  
  [thiagohmcruz](https://github.com/thiagohmcruz)
```

</p>

lib/xcodeproj/scheme/execution_action.rb
```diff
@@ -0,0 +1,99 @@
...
+      end
+
+      # @return [String]
+      #         The ActionType for this ExecutionAction.
+      #         One of two values:
+      #
+      #         Xcode.IDEStandardExecutionActionsCore.ExecutionActionType.ShellScriptAction,
+      #         Xcode.IDEStandardExecutionActionsCore.ExecutionActionType.SendEmailAction
+      #
+      def action_type?
```
amorde, line 21: the `?` suffix on an accessor should only be used if the return type is `Bool`

lib/xcodeproj/scheme/execution_action.rb
```diff
@@ -0,0 +1,99 @@
...
+        elsif @action_content.is_a?(SendEmailActionContent)
+          type == Constants::EXECUTION_ACTION_TYPE[:send_email_action]
+        else
+          raise "[Xcodeproj] Invalid ActionType `#{type}` for ActionContent `#{@action_content.class}`"
+        end
+      end
+
+      # @param [ShellScriptActionContent, SendEmailActionContent] value
+      #        Checks if value matches the expected action_type if present.
+      #
```
amorde, line 84: nit: add a `@return [Bool] true if the value is valid`

lib/xcodeproj/scheme/send_email_action_content.rb
```diff
@@ -0,0 +1,82 @@
+module Xcodeproj
+  class XCScheme
+    # This class wraps a 'ActionContent' node of type
+    # 'Xcode.IDEStandardExecutionActionsCore.ExecutionActionType.SendEmailAction' of a .xcscheme XML file
+    #
+    class SendEmailActionContent < XMLElementWrapper
+      # @return [Bool]
+      #         Wheter or not this action should attach log to email
```
amorde, line 8: `wheter` --> `whether`

lib/xcodeproj/scheme/send_email_action_content.rb
```diff
@@ -0,0 +1,82 @@
...
+      #
+      attr_reader :attach_log_to_email
```
amorde, line 10: this should get an alias `attach_log_to_email?`

lib/xcodeproj/scheme/send_email_action_content.rb
```diff
@@ -0,0 +1,82 @@
...
+      #        Set the email recipient of this ActionContent
+      #
+      def email_recipient=(value)
+        @xml_element.attributes['emailRecipient'] = value
+      end
+
+      # @return [String]
+      #         The email subject of this ActionContent
+      #
+      def email_subject?
```
dnkoutso, line 56: similarly for all of these only use `?` in method names if the return type is `BOOL` which is idiomatic Ruby.

CHANGELOG.md
```diff
@@ -4,7 +4,9 @@
 
 ##### Enhancements
 
-* None.  
+* Add support for pre/post-actions in scheme actions
```
dnkoutso, line 7: nit: add two empty spaces at the end of this for markdown formatting.

lib/xcodeproj/scheme/abstract_scheme_action.rb
```diff
@@ -23,6 +23,102 @@ def build_configuration
...
+
+      #-------------------------------------------------------------------------#
+
+      private
+
+      # @!group Private helpers
+
+      # @return [Bool]
+      #         The AnalyzeAction does not support pre/post actions (see in Xcode)
+      def pre_and_post_actions_disabled
```
dnkoutso, line 119: nits: add an empty line after the ruby doc and the method name.

This one can be renamed to `pre_and_post_actions_disabled?`

lib/xcodeproj/scheme/execution_action.rb
```diff
@@ -0,0 +1,89 @@
...
+
+      # @!group Private helpers
+
+      # @return [Bool]
+      #         True if value is valid
+      #
+      # @param [ShellScriptActionContent, SendEmailActionContent] value
+      #        Checks if value matches the expected action_type if present.
+      #
+      def validate_action_content(value)
```
dnkoutso, line 77: maybe rename `valid_action_content?`?

dnkoutso (approved):

lib/xcodeproj/constants.rb
```diff
@@ -457,8 +457,8 @@ module Constants
     # @return [Hash] Possible types for a scheme's 'ExecutionAction' node
     #
     EXECUTION_ACTION_TYPE = {
-      :shell_script_action  => 'Xcode.IDEStandardExecutionActionsCore.ExecutionActionType.ShellScriptAction',
-      :send_email_action    => 'Xcode.IDEStandardExecutionActionsCore.ExecutionActionType.SendEmailAction',
+      :shell_script  => 'Xcode.IDEStandardExecutionActionsCore.ExecutionActionType.ShellScriptAction',
```
segiddins, line 460: is changing these keys a breaking change?

lib/xcodeproj/scheme/abstract_scheme_action.rb
```diff
@@ -51,29 +56,37 @@ def pre_actions=(pre_actions)
       end
 
       # @param [ExecutionAction] pre_action
-      #        The ExecutionAction to add to the PreActions node associated with this action
+      #        Add an action to the list of actions to run before this scheme action.
+      #        It can be either a 'Run Script' or a 'Send Email' action.
       #
       def add_pre_action(pre_action)
+        return if pre_and_post_actions_disabled
```
segiddins, line 63: doing this check here seems weird -- why is it not valid to access the pre / post actions, even if Xcode won't run them?

lib/xcodeproj/scheme/execution_action.rb
```diff
@@ -3,56 +3,60 @@ class XCScheme
...
+      #        Xcode.IDEStandardExecutionActionsCore.ExecutionActionType.ShellScriptAction,
+      #        Xcode.IDEStandardExecutionActionsCore.ExecutionActionType.SendEmailAction
+      #
       # @param [REXML::Element] node
       #        The 'ExecutionAction' XML node that this object will wrap.
       #        If nil, will create an empty one
       #
-      def initialize(node = nil)
-        create_xml_element_with_fallback(node, 'ExecutionAction')
+      def initialize(action_type = nil, node = nil)
```
segiddins, line 16: this is a breaking change -- the action type either needs to come after the node or be a kwarg

lib/xcodeproj/scheme/execution_action.rb
```diff
@@ -3,56 +3,60 @@ class XCScheme
     # This class wraps the ExecutionAction node of a .xcscheme XML file
     #
     class ExecutionAction < XMLElementWrapper
+      # @param [String] action_type
```
segiddins, line 6: why do we not prefer passing the action type as the symbol used in the constants hash?

lib/xcodeproj/scheme/execution_action.rb
thiagohmcruz: In my mind I keep associating `?` with `Optional<>` and not `Bool`, my bad!

lib/xcodeproj/constants.rb
thiagohmcruz: `EXECUTION_ACTION_TYPE ` is new and this is just me changing my mind on how to name these symbols while implementing this. So I don't think isn't a breaking change unless I'm missing something.

lib/xcodeproj/scheme/abstract_scheme_action.rb
thiagohmcruz: I was trying to support only what has a purpose in Xcode, I wanted to avoid putting things in the `.xcscheme` file that are not gonna be used and maybe one day cause a weird Xcode crash? Maybe I'm being overly precautious though?

lib/xcodeproj/scheme/execution_action.rb
thiagohmcruz: I'll put `action_type` after the `node` 👍 but sorry can you clarify where this would break? This is a new class but I'm wondering now if somewhere else `node` is expected to always be the first argument?

lib/xcodeproj/scheme/execution_action.rb
thiagohmcruz: I missed that, makes sense to me 👍

thiagohmcruz:
@segiddins push a few changes, thx for catching those!

lib/xcodeproj/scheme/abstract_scheme_action.rb
thiagohmcruz: Thoughts here? Maybe I'm missing something 🤔

lib/xcodeproj/scheme/execution_action.rb
segiddins: ah sorry in the diff it looked like a change, not net-new code

lib/xcodeproj/scheme/abstract_scheme_action.rb
segiddins: I think we should still allow accessing and modifying these attributes, regardless of whether they're disabled. It's up to the caller to determine if they want behavior conditional on things being enabled or not

thiagohmcruz: Done!

tinder-owenthomas:
@dnkoutso @amorde I'd love to be able to use post-actions directly without manipulating the XML manually (because that's causing complications with Xcode 12). Any chance a new release with this functionality may be coming soon? 

Thank you! 🙏

PR #807, opened 2021-01-30
Adds the ability to specify sorting options for groups.

Zanchee:
The `sort` command doesn't allow the user to specify how they want their project sorted. Instead, groups and files are sorted alphabetically and interleaved with each other.

Xcodeproj already supports the ability to sort with groups above or below files, but the functionality is not exposed to the cli command.

This PR simply exposes that functionality with  a new option.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,9 @@
 
 ##### Enhancements
 
-* Add support for group options when using  the sort command
+* Add support for group options when using the sort command
   [zanchee](https://github.com/Zanchee)
+  [imachumphries](https://github.com/imachumphries)
   [#807](https://github.com/CocoaPods/Xcodeproj/pull/807)
 
 * Add support for pre/post-actions in scheme actions  
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 ##### Enhancements
 
+* Add support for group options when using  the sort command
+  [zanchee](https://github.com/Zanchee)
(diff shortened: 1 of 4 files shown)


dnkoutso:
Thanks! can you add a changelog entry and some tests?

Zanchee:
> Thanks! can you add a changelog entry and some tests?

I was looking around the repo at some example tests, and it looks like there is decent coverage for the sort functionality already. Is there an example I can look at for testing the actual command options and such? My only changes are command file.

dnkoutso:
Ah looks like sort.rb does not have any specs...there should be `sort_spec.rb`....

Zanchee:
> Ah looks like sort.rb does not have any specs...there should be `sort_spec.rb`....

What would the objectives of these tests be? (Sorry, I'm new to ruby) I can see that the `group_spec.rb` already tests both the `above` and `below` functionality. 

I tried looking for examples of specs for any of the commands  under `lib/xcodeproj/commands` but I didn't see a spec for any of the CLI commands.

dnkoutso:
The new tests should test the fact that anyone can pass in the new parameters you added in `sort.rb` and make sure the correct parameters are delegated further down.

Zanchee:
Might take a while for me to investigate how to write these tests. I have 0 Ruby experience, so if anyone else wants to contribute, that would be amazing. Otherwise, I'll get around to it in my free time.

Zanchee:
@dnkoutso Managed to get a friend with some Ruby knowledge to add some tests. 🥳

segiddins (approved):

PR #820, opened 2021-05-26
Add support to parse OTHER_LDFLAGS arg files using '@'

dnkoutso:


segiddins (approved):

spec/config/other_linker_flags_parser_spec.rb
```diff
@@ -123,12 +169,15 @@
       flags << '-lxml2.2.7.3'
       flags << '-l "Pods-AFNetworking iOS Example-AFNetworking"'
       flags << '-l"Pods-AFNetworking iOS Example-AFNetworking"'
+      flags << '@"${PODS_ROOT}/Target Support Files/other_ldflags"'
+      flags << '@ "${PODS_ROOT}/Target Support Files/other_ldflags"'
```
amorde, line 173: is there a test case covering a space between `@` and the quoted file path?

amorde (approved):

spec/config/other_linker_flags_parser_spec.rb
dnkoutso: I believe so yes, right above

amorde: thanks, I missed it

PR #832, opened 2021-07-13
Add fallback to sort group by full path

igor-makarov:
This PR resolves an issue where having several items that have the same `basename` and `extname` would be sorted in an unstable manner.

Example:
```
PBXVariantGroup
|
+- en.lproj/InfoPlist.strings
+- he.lproj/InfoPlist.strings
+- it.lproj/InfoPlist.strings
+- zh-Hant.lproj/InfoPlist.strings
```

Because essentially all of these files have the same `basename` and `extname`, the sort becomes non-deterministic.

Adding a fallback by full path, eg.g `en.lproj/InfoPlist.strings` resolves this.

### DIFF
--- a/lib/xcodeproj/project/object/group.rb
+++ b/lib/xcodeproj/project/object/group.rb
@@ -440,10 +440,12 @@ module Xcodeproj
 
             result = File.basename(x.display_name.downcase, '.*') <=> File.basename(y.display_name.downcase, '.*')
             if result.zero?
-              File.extname(x.display_name.downcase) <=> File.extname(y.display_name.downcase)
-            else
-              result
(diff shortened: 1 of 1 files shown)


dnkoutso:
any quick test to add for this?

igor-makarov:
@dnkoutso yeah, I was going to - but where are the test runs? 😳

dnkoutso:
uuuuh, not sure? we need to fix that.

igor-makarov:
@dnkoutso move it to GH actions perhaps?

dnkoutso:
yes sir that would be best. I might have time in a day or two to take care of it!

igor-makarov:
Opened #835 instead, sorry about the mess.

PR #835, opened 2021-07-14
Add fallback to sort group by full path

igor-makarov:
This PR resolves an issue where having several items that have the same `basename` and `extname` would be sorted in an unstable manner.

Example:
```
PBXVariantGroup
|
+- en.lproj/InfoPlist.strings
+- he.lproj/InfoPlist.strings
+- it.lproj/InfoPlist.strings
+- zh-Hant.lproj/InfoPlist.strings
```

Because essentially all of these files have the same `basename` and `extname`, the sort becomes non-deterministic.

Adding a fallback by full path, eg.g `en.lproj/InfoPlist.strings` resolves this.


dnkoutso:
amazing!

dnkoutso:
Still would love a test here perhaps, merging.

PR #839, opened 2021-07-16
Add an unambiguous sort order to build phase files

igor-makarov:
This PR resolves an issue where build files of the same build phase sharing the same filename were being treated as equal by the sort algorithm. This resulted in an unstable sort order.

The PR introduces a defined sort order for build files that aims to remove almost all ambiguity:
* Sort by display base name, lowercased
* If ambiguous, sort by display name, lowercased
* If ambiguous, sort by full path, lowercased

To be clear - before the PR, the build files were being sorted by the default sort impl in `AbstractObject`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,13 @@
 
 ##### Enhancements
 
-* None.  
+* Add an unambiguous sort order to build phase files.  
+  [Igor Makarov](https://github.com/igor-makarov)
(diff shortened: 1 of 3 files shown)


segiddins (approved):
Nice! Changelog entry?

igor-makarov:
Yeah, turns out I forgot to add one for the previous PR too! Added both now.

dnkoutso (approved):

PR #851, opened 2021-08-04
Fix crashing subprojects

avetiso:
Picking up where [#762](https://github.com/CocoaPods/Xcodeproj/pull/762) left off. Updated the specs to show what the new remote_info actually is. Fixed some whitespacing.


CHANGELOG.md
```diff
@@ -138,6 +138,10 @@
   [nickgravelyn](https://github.com/nickgravelyn)
   [#757](https://github.com/CocoaPods/Xcodeproj/pull/757)
 
+* Fix crash when adding sub project.
```
dnkoutso, line 141: nit: please add two empty spaces after the period for markdown formatting.

also add your name for credit!

CHANGELOG.md
```diff
@@ -138,6 +138,10 @@
...
+  [HDB-Li](https://github.com/HDB-Li)
+  [#762](https://github.com/CocoaPods/Xcodeproj/pull/762)
```
dnkoutso, line 143: nit: this should be 851

lib/xcodeproj/project/object/helpers/file_references_factory.rb
```diff
@@ -180,15 +180,16 @@ def new_subproject(group, path, source_tree)
             ref = new_file_reference(group, path, source_tree)
             ref.include_in_index = nil
 
-            product_group_ref = find_products_group_ref(group, true)
+            product_group_ref = group.project.new(PBXGroup)
```
dnkoutso, line 183: should we create a new group each time?

CHANGELOG.md
avetiso: Sorry, modified before I saw these comments.

lib/xcodeproj/project/object/helpers/file_references_factory.rb
avetiso: Waiting to hear back from a team member regarding our Xcode set up. Will get back to you as soon as I know.

CHANGELOG.md
avetiso: Fixed in update.

lib/xcodeproj/project/object/helpers/file_references_factory.rb
```diff
@@ -237,6 +238,24 @@ def find_products_group_ref(group, should_create = false)
...
+          #
+          # @param  [Project] project
+          #         The project to which to find the target.
+          #
+          # @note   If the target is found, then return the name of target,
+          #         otherwise, the default value 'Subproject' will be returned
+          #
+          # @return [String] The remote info name.
+          #
+          def find_remote_info_name_for_product_reference(ref, project)
```
dnkoutso, line 254: I think you can re-write this as:

```ruby
          def find_remote_info_name_for_product_reference(ref, project)
            project.native_targets.find { |t| t.product_reference == ref }&.name || 'Subproject'
          end          
```

avetiso: RuboCop complains over this for using the 2.1 parser. Want me to update to 2.3?

avetiso: It complains about the ampersand. Changing to 2.3 complains about frozen string literals and freezing immutable objects and other stuff.

lib/xcodeproj/project/object/helpers/file_references_factory.rb
avetiso: It works the same for us. The concern we had with having this be a new group was that our subprojects would generate as a new subproject folder instead of a new subproject xcodeproj file in the navigator. I just ran through this with our main engineer that set up our Xcode project and verified that it still sets up a new <subproject>.xcodeproj file as we want.

As far as side effects however, I don't believe him or I can really speak beyond just our project. If you have any other people that are very familiar with Xcode that want to take a look at this I'd be happy to see the input and correct anything that's wrong, but that's out of my expertise at that point.

I just know that this fixes our crashing when trying to change our project settings. 

Happy to hear any concerns and implement any changes.

Also, I found this article that describes a much more verbose fix for the same issue. Take a look and compare and let me know what you think, perhaps their solution is better? https://www.programmersought.com/article/83996170465/

avetiso: Any feedback? Sorry we can't be of more help.

avetiso: We appreciate your time. Totally understand not wanting to ship a breaking change.

avetiso: Hey there, don't mean to rush you if you're busy, just checking in if you had any updated thoughts on this.

avetiso: Let us know if you'll find the time to get to this anytime soon please @dnkoutso 🙂 If not it make more sense for us to fork and just wait for this to be fixed before reverting back to using the official version.

dnkoutso: Sorry for the long delay, I recommend for now to fork and proceed. I have not had time to review this and evaluate it. I apologize but normal day-to-day work stuff................

avetiso: No worries, totally get it. We will patch and host it locally for now, and when you get to this and check it out, it would be great if you could let us know so we can revert back to the public one. :)

avetiso: Just FYI, after doing more of our own internal testing it looks like this ends up breaking one of the entries in our linking phase.

dnkoutso: thats "good" to hear I guess! Without some deeper investigation I would be afraid to land this. Again really sorry for lack of traction here, the repercussions can be large.

avetiso: Yeah, totally. Wouldn't want to break anything for anyone. 

Looking around though, it does seem like this has been an issue for multiple people. Are there any plans to investigate this further? Or, perhaps, to write up a guide on how to properly create subprojects so that we can see if perhaps we are mis-using the library?

Issue #797, opened 2020-10-29
Schemes issue


yakimant:
```
Xcodeproj::Workspace.new_from_xcworkspace("My.xcworkspace").schemes
Xcodeproj::Project.schemes("My.xcodeproj")
```

are showing much less schemes, than
```
xcodebuild -list -workspace My.xcworkspace
xcodebuild -list -project My.xcodeproj
```

dive:
Seems like the same problem as #607 and #557.

rogerluan:
@yakimant could you provide a zip file containing an Xcode project that can reproduce this issue? 
Also knowing your xcodeproj version and xcodebuild version should help 🙏

rogerluan:
I recommend trying this solution, as it seems to have worked for them: https://github.com/fastlane/fastlane/issues/17521#issuecomment-722154367

dnkoutso:
A PR to fix this would be welcome!

dongdonggaui:
hi @dnkoutso , could you take a look at this pull request? https://github.com/CocoaPods/Xcodeproj/pull/869

PR #869, opened 2021-12-31
Fix workspace schemes doesn't match the result of `xcodebuild -list -workspace` #797

dongdonggaui:


dongdonggaui:
fix https://github.com/CocoaPods/Xcodeproj/issues/797

dnkoutso:
@dongdonggaui can you please add a test case? thank you!

dongdonggaui:
@dnkoutso got it，thanks for the reply

dnkoutso:
ping @dongdonggaui if possible please add a test

lib/xcodeproj/workspace.rb
```diff
@@ -82,6 +82,14 @@ def self.new_from_xcworkspace(path)
     def self.from_s(xml, workspace_path = '')
       document = REXML::Document.new(xml)
       instance = new(document)
+      # Load projects reference in "Pods/" if needed
+      pods_path = "#{File.expand_path("..", workspace_path)}/Pods"
```
dnkoutso, line 86: I think this project doesnt know about "Pods". This fix might be better in CocoaPods gem instead.

kaporn1990:
sudo gem install cocoapods

PR #873, opened 2022-02-25
Create command

iamleeg:
We want to add to the command set of `xcodeproj` so that we can manipulate Xcode projects on non-Mac platforms. The first command is `xcodeproj create PROJECT`.

### DIFF
--- a/lib/xcodeproj/command.rb
+++ b/lib/xcodeproj/command.rb
@@ -3,6 +3,7 @@ module Xcodeproj
   require 'claide'
 
   class Command < CLAide::Command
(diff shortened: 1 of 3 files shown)


igor-makarov:
@iamleeg Have you taken a look at [Xcake](https://github.com/igor-makarov/xcake)? It's a fully featured project generator with an easy DSL that might already do what you need. 

It uses Xcodeproj under the hood.

iamleeg:
I've been using Xcake for the last month and it's sufficient for my needs. I leave it up to you whether you merge or reject this PR.

PR #877, opened 2022-04-19
Bump SDK, Swift, and upgrade constants

sethfri:
_(Reviewable commit-by-commit)_

* Bump last known SDK versions
* Bump Swift version to 5.6
* Bump last upgrade version to 1330


lib/xcodeproj/constants.rb
```diff
@@ -4,27 +4,27 @@ module Xcodeproj
...
+    LAST_KNOWN_WATCHOS_SDK = '8.5'
 
     # @return [String] The last known archive version to Xcodeproj.
     #
     LAST_KNOWN_ARCHIVE_VERSION = 1
 
     # @return [String] The last known Swift version (stable).
     #
-    LAST_KNOWN_SWIFT_VERSION = '5.0'
+    LAST_KNOWN_SWIFT_VERSION = '5.6'
```
sethfri, line 27: This might not actually be right? In Xcode, I see Swift 4, Swift 4.2, and Swift 5 as options, but nothing beyond Swift 5. I guess minor versions of Swift don't have their own compiler modes?

segiddins: I think this stays at 5 now, swift 4.2 was the last minor version to get its own compiler mode

lib/xcodeproj/constants.rb
```diff
@@ -399,7 +399,7 @@ module Constants
         'GCC_WARN_UNUSED_VARIABLE'                => 'YES',
         'MTL_FAST_MATH'                           => 'YES',
         'PRODUCT_NAME'                            => '$(TARGET_NAME)',
-        'SWIFT_VERSION'                           => '5.0',
+        'SWIFT_VERSION'                           => '5.6',
```
segiddins, line 402: likewise, this should stay at 5

lib/xcodeproj/constants.rb
amorde: yeah this should stay at 5.0 for now

PR #915, opened 2023-07-14
target_dependency: Fix crash if `target_proxy` is `nil`.

byohay:
If `target_proxy` is `nil`, calling `to_tree_hash` on it will raise an
exception: undefined method `to_tree_hash' for nil:NilClass
(NoMethodError).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,9 @@
 
 ##### Bug Fixes
 
-* None.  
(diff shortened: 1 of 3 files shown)


dnkoutso:
Can you provide more info about the failure case? Also what about a test?

byohay:
@dnkoutso sorry, added a test and explanation.

amorde (approved):

PR #919, opened 2023-09-11
Use relative_path for local Swift packages

jsorge:
PR #911 added support for local packages but did not add support for the `relativePath` property inside of the local package definition. Here's a sample from my pbxproj file:

```
/* Begin XCLocalSwiftPackageReference section */
		93ED17092AAED02600B3DB25 /* XCLocalSwiftPackageReference "../../Packages/SharedResources" */ = {
			isa = XCLocalSwiftPackageReference;
			relativePath = ../../Packages/SharedResources;
		};
```
When I try to use the new local package model I got the following error: `[!] Xcodeproj doesn't know about the following attributes {"relativePath"=>"../../Packages/SharedResources"} for the 'XCLocalSwiftPackageReference' isa.` and I believe that was because the class was looking for `path` (which doesn't exist) and not `relativePath` which does.

This PR fixes that problem for local packages.

### DIFF
--- a/lib/xcodeproj/project/object/swift_package_local_reference.rb
+++ b/lib/xcodeproj/project/object/swift_package_local_reference.rb
@@ -21,7 +21,7 @@ module Xcodeproj
         # @return [String] the path of the local Swift package reference.
         #
         def display_name
-          return path if path
+          return relative_path if relative_path
           super
         end
       end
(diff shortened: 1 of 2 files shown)


dnkoutso:
@jsorge can you have both `path` and `relativePath`? Why is the first one replaced?

jsorge:
> @jsorge can you have both `path` and `relativePath`? Why is the first one replaced?

I replaced it because I didn’t see `path` as a thing that could be parsed from the project file. Happy to add it back if it could be there.

dnkoutso:
Wonder if an Xcode version upgrade replaced it or added it...................also wonder if there is harm in keeping both

jsorge:
Added it back @dnkoutso

serbancoroiu:
Fixed in the last release, works now. Thanks!

Issue #864, opened 2021-11-12
Unexcepted additional .git keyword in XCRemoteSwiftPackageReference


powerwolf543:
Xcodeproj version: `1.20.0`
I want to update the version of a swift package. Hence I emulated the objects and found the package. Then I modified the version and saved it. But I couldn't open my Xcode project. It seems my .pbxproj format is incorrect.

<img width="656" alt="Screen Shot 2021-11-12 at 3 39 32 PM" src="https://user-images.githubusercontent.com/16394562/141428792-7688310e-9841-4760-a587-2cb133706566.png">

Here is my code:
``` ruby
project_path = './myproject.xcodeproj'

if !File.exist?(project_path) 
    print "Didn't find .xcodeproj file at '#{project_path}'\n"
    exit(1)
end

project = Xcodeproj::Project.open(project_path)

project.objects.each do |object|
    next unless 'some_package' == object.display_name && object.is_a?(Xcodeproj::Project::Object::XCSwiftPackageProductDependency)
    object.package.requirement["version"] = '456'
    print "display_name: #{object.display_name}, requirement: #{object.package.requirement["version"]}\n"
end

project.save
```

zachwaugh:
I'm seeing this as well, but I'm not modifying any Swift package related fields. I'm just updating the `MARKETING_VERSION` and saving, and then all my dependencies have the `.git` appended. I'm using Xcodeproj `1.21.0` and Xcode `13.1`. This didn't seem to mess up the project file or prevent me from opening, but would hope that unrelated properties wouldn't be modified during save.

```ruby
project.build_configurations.each do |configuration|
  configuration.build_settings["MARKETING_VERSION"] = new_version
end

project.save
```

mccarron:
I just figured it out, as it was only happening to some of our packages, not all of them. 

It was only happening to projects with `.git` at the end of the `XCRemoteSwiftPackageReference` `repositoryURL`. So to avoid this going forward, I removed the`.git` from those URLs. Ran `pod install` again and did an Xcode build. Now all is well between both tools with the naming of those references.

Hopefully this helps someone else.

Top-Pattarapol:
do you have any solution for this?, I found same issue🥺

PR #921, opened 2023-09-28
Update remote Swift packages annotations to match Xcode

alexandre-pod:
I am using two tools that use `Xcodeproj` to manipulate Xcode files ([ccios](https://github.com/faberNovel/ccios) and [Kintsugi](https://github.com/Lightricks/Kintsugi/)). But on projects using Swift packages the `Xcodeproj` updates  is not matching what Xcode 14 and 15 generates, it has additional ".git" in packages annotations. 

For example, with a clean project with a single Swift package dependency, when I just read and save the Xcode project using this code (`Xcodeproj::Project.open('./SimpleApp.xcodeproj').save('./SimpleApp.xcodeproj')`), this happens in git:

diff of the .xcodeproj file showing the Xcodeproj adding .git extension in annotations

This PR make `Xcodeproj` match the Xcode behavior

### DIFF
--- a/lib/xcodeproj/project/object/swift_package_remote_reference.rb
+++ b/lib/xcodeproj/project/object/swift_package_remote_reference.rb
@@ -18,7 +18,7 @@ module Xcodeproj
         #--------------------------------------#
 
         def ascii_plist_annotation
-          " #{isa} \"#{File.basename(display_name)}\" "
+          " #{isa} \"#{File.basename(display_name,".git")}\" "
         end
 
         # @return [String] the name of the remote Swift package reference.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,9 @@
 
 ##### Bug Fixes
 
-* None.  
+* Update remote Swift packages annotations to match Xcode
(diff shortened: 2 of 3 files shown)


lib/xcodeproj/project/object/swift_package_remote_reference.rb
```diff
@@ -18,7 +18,7 @@ class XCRemoteSwiftPackageReference < AbstractObject
         #--------------------------------------#
 
         def ascii_plist_annotation
-          " #{isa} \"#{File.basename(display_name)}\" "
+          " #{isa} \"#{File.basename(display_name,".git")}\" "
```
amorde, line 21: I think this might be the inverse here. If I make a fresh project in Xcode, add SPM dependencies and commit it, the `.git` is not there.

Then I open and save the project like `Xcodeproj::Project.open(project_path).save` and it (seemingly incorrectly) adds the `.git` suffix. This is covered in https://github.com/CocoaPods/Xcodeproj/issues/864

amorde:
Able to reproduce, and found https://github.com/CocoaPods/Xcodeproj/issues/864 which is related. Thanks!

Issue #906, opened 2023-03-29
[!] [!] Xcodeproj doesn't know about the following attributes {"productname"=>"MinaretTests"} for the 'PBXNativeTarget' isa.


SamirWow:

appfrilans:
Same issue here! This after the new Xcode 15 release. I guess I need to downgrade!?!?

[!] Xcodeproj doesn't know about the following attributes {"expectedSignature"=>"AppleDeveloperProgram:Y28xxxSHX7:Stripe Inc"} for the 'PBXFileReference' isa.
If this attribute was generated by Xcode please file an issue: https://github.com/CocoaPods/Xcodeproj/issues/new

dnkoutso:
This needs a PR to teach Xcodeproj about this new attribute.

gsbernstein:
Same with Validic's SDK

```
[!] [!] Xcodeproj doesn't know about the following attributes {"expectedSignature"=>"AppleDeveloperProgram:24YV5MVWU8:Validic, Inc."} for the 'PBXFileReference' isa.
If this attribute was generated by Xcode please file an issue: https://github.com/CocoaPods/Xcodeproj/issues/new

[!] [!] Xcodeproj doesn't know about the following attributes {"expectedSignature"=>"AppleDeveloperProgram:24YV5MVWU8:Validic, Inc."} for the 'PBXFileReference' isa.
If this attribute was generated by Xcode please file an issue: https://github.com/CocoaPods/Xcodeproj/issues/new

[!] [!] Xcodeproj doesn't know about the following attributes {"expectedSignature"=>"AppleDeveloperProgram:24YV5MVWU8:Validic, Inc."} for the 'PBXFileReference' isa.
If this attribute was generated by Xcode please file an issue: https://github.com/CocoaPods/Xcodeproj/issues/new

[!] [!] Xcodeproj doesn't know about the following attributes {"expectedSignature"=>"AppleDeveloperProgram:24YV5MVWU8:Validic, Inc."} for the 'PBXFileReference' isa.
If this attribute was generated by Xcode please file an issue: https://github.com/CocoaPods/Xcodeproj/issues/new
```

The attribute was generated by xcode after I updated the frameworks.

Viswanth92:
Hi I am noticing the issue still after updating to the latest cocoapods 1.15.2. Is this resolved?

ChrSngular:
I have the latest 1.15.2 and still having this issue, some ETA?

PR #924, opened 2023-11-03
Add `expected_signature` support for `PBXFileReference`

hieu9102002:
Trivially adds Expected signature support to PBX File Reference. This option is apparently only available to `xcframeworks` references.

Documentation: https://developer.apple.com/documentation/Xcode/verifying-the-origin-of-your-xcframeworks

Mentioned in #906 with a real example.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ##### Enhancements
 
-* Add `expected_signature support` for `PBXFileReference`
+* Add `expected_signature support` for `PBXFileReference`.  
   [hieu9102002](https://github.com/hieu9102002)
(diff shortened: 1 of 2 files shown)


hieu9102002:
cc: @dnkoutso , just wanted to loop you in since you are a maintainer of this project.

rmmmmm41:
@hieu9102002 Hello, I have encountered this issue too. I want to ask if there will be any serious consequences if we ignore this bug.

rmmmmm41:
During the build process, my program currently seems to have no abnormalities

Footpad:
@dnkoutso could we get this merged? Xcode 15 is now expecting to be able to have these signatures and it's blocking our builds.

PR #932, opened 2024-02-22
swift_package_product_dependency: override `ascii_plist_annotation`.

GalBerezansky:
We have a build plugin which is a Swift package, it's represented as a `XCSwiftPackageProductDependency` object. The display name contains a "plugin:" prefix so when it's serialized it's everywhere. The problem is that Xcode removes this prefix and uses the package name in the keys of the serialized objects and in refs to them.

Below is an example before and after the change.

before (wrong):
```
/* Begin XCSwiftPackageProductDependency section */
		19AA76E52A9F557900F1F6A1 /* plugin:SomePackageName */ = {
			isa = XCSwiftPackageProductDependency;
			productName = "plugin:SomePackageName";
		};
```
after (correct):
```
/* Begin XCSwiftPackageProductDependency section */
		19AA76E52A9F557900F1F6A1 /* SomePackageName */ = {
			isa = XCSwiftPackageProductDependency;
			productName = "plugin:SomePackageName";
		};
```

### DIFF
--- a/lib/xcodeproj/project/object/swift_package_product_dependency.rb
+++ b/lib/xcodeproj/project/object/swift_package_product_dependency.rb
@@ -23,6 +23,15 @@ module Xcodeproj
           return product_name if product_name
           super
         end
+
+        # @!group AbstractObject Hooks
+        #--------------------------------------#
+        #
+        # @return [String] the asciii plist annotation of the Swift package product dependency.
+        #
+        def ascii_plist_annotation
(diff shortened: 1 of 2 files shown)


GalBerezansky:
The `.delete_prefix('plugin:')` is a little meh but I couldn't find a better solution for this issue.

GalBerezansky:
@dnkoutso Can you take a look please? Thank you 😄

PR #940, opened 2024-06-04
fix: Support optional includes

Werelds:
This PR adds support for the optional include syntax `#include?`. Xcode will include these files if they exist and ignore the entries otherwise.

Due to how the library currently handles includes, it's only the syntax that needs to be supported as it already considers all includes optional.


segiddins:
Needs test coverage added

Werelds:
Added an explicit test for it.

segiddins (approved):

Issue #933, opened 2024-03-01
Xcode 15.3 Compatibility


Brett-Best:
Xcode 15.3 introduces:
- object version 63
- new property attributes {"minimizedProjectReferenceProxies"=>"1"} for the 'PBXProject' isa

wojciech-kulik:
I noticed also another problem. When Xcode 15.3 is opened and I do any changes to the project using `Xcodeproj`. The `App.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved` gets deleted every time.

If I do the same with closed Xcode and open it after changes, it's fine. It also works correctly with Xcode 15.2.

wojciech-kulik:
It turns out that most likely it's not a Xcodeproj-related bug. Whenever you change `xcodeproj` outside of Xcode when it's opened, this issue occurs.

Whirlwind:
@amorde Could you have a look this issue?

Issue #947, opened 2024-07-05
Update REXML to fix DoS Vulnerability (CVE-2024-35176)


fchiusolo:
Hi team,

There's a DoS vulnerability in `rexml` before version 3.2.7. It affects `xcodeproj` through `fastlane`. Can you update `rexml` to version 3.2.7 or later?

More details: [ruby-lang.org](https://www.ruby-lang.org/en/news/2024/05/16/dos-rexml-cve-2024-35176/).

Thanks!

bmedenwald:
It's worse now. New CVE: https://www.ruby-lang.org/en/news/2024/07/16/dos-rexml-cve-2024-39908/

This needs to be resolved.

bmedenwald:
Added PR: https://github.com/CocoaPods/Xcodeproj/pull/948

samfranz:
When can we expect a release with this change?

elkinjosetm:
The PR was already merged, any idea when should we expect to have a new release?

sphanley:
Seconding the request for a release at the earliest convenience– my enterprise team is running into security warnings related to this rexml dependency, so it would be hugely helpful to have a new version released with the recently updated constraint.

sphanley:
For others impacted by these security warnings in the short term, if you're comfortable being "between releases", you can target the current commit by adding this to your Gemfile: 

> gem "xcodeproj",  git: "https://github.com/CocoaPods/Xcodeproj.git", :branch => "master", ref: "fe55cf5"

Since the Version defined in the repo is still 1.24.0, this will let you consume the updated dependency list while still satisfying transitive dependencies on 1.24.0.

kconner:
Please release a numbered version with the merged dependency update. The latest version of this library, 1.24.0, is still vulnerable.

konrad-gibaszewski:
Please release a numbered version with the merged dependency update.

0rax:
For anybody still struggling with this, you can point your `Gemfile` to this git repository directly to retrieve the latest version from master.

This worked fine in my case:

```ruby
source 'https://rubygems.org'

ruby '>= 2.6.10'

gem 'cocoapods', '>= 1.15.2'
gem 'xcodeproj', '~> 1.24', git: 'https://github.com/CocoaPods/Xcodeproj.git'
gem "rexml", "~> 3.3.2"
```

Once a new version is release, just remove the `git:` part and update the version identifier next to `xcodeproj`.

[...]

PR #953, opened 2024-08-02
Don't raise an exception on newer Xcode than latest known

Kaspik:
- remove exception on newer Xcode version and log only warning to inform user they are using newer xcode than the latest supported by xcodeproj - if things don't work, it will fail automatically on the right place - but chance is newer Xcode simply works so there is no reason why users should be blocked by outdated xcodeproj

Fixes issues like https://github.com/CocoaPods/Xcodeproj/issues/933 where [fix](https://github.com/CocoaPods/Xcodeproj/pull/934) is merged in but **still not released** and users are blocked for months.

### DIFF
--- a/lib/xcodeproj/project.rb
+++ b/lib/xcodeproj/project.rb
@@ -221,11 +221,11 @@ module Xcodeproj
       end
 
       if archive_version.to_i > Constants::LAST_KNOWN_ARCHIVE_VERSION
-        puts "WARN: [Xcodeproj] Unknown archive version (#{archive_version.to_i})."
+        puts "WARN: [Xcodeproj] Archive version (#{archive_version.to_i}) is higher than the latest supported by xcodeproj (#{Constants::LAST_KNOWN_ARCHIVE_VERSION})."
       end
 
       if object_version.to_i > Constants::LAST_KNOWN_OBJECT_VERSION
-        puts "WARN: [Xcodeproj] Unknown object version (#{object_version.to_i})."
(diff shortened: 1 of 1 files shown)


martinpucik (approved):
This works for issue we're facing 👍

orta:
Can you switch this around to a warning please, and I'll get this merged - I think it's the right thing to do at this point

orta:
There's also the one above, which might need it too?

Kaspik:
Happy to do so! (once I find how I throw warning 😃 )

Kaspik:
Update to puts with a `WARN:` prefix

orta:
We use `UI.warn` - https://github.com/CocoaPods/Xcodeproj/blob/fe55cf57badef2ca27fb9d4f801d9b834de1f871/lib/xcodeproj/project/object.rb#L321C13-L321C22 which should make it consistent with other stuff (I'd offer a suggestion but there could be lint issues from it )

Kaspik:
Thanks, done! ✅

Kaspik:
@orta Once this gets out, any chance we could do a release? Not only for this, but the [CVE-Vulnerability](https://github.com/CocoaPods/Xcodeproj/issues/947) fix should also get out finally. :)

orta:
That is more complex and out of my hands to commit to I'm afraid, but I'd like to see it happen too

Kaspik:
Understand, let's merge this and we will see how others feel about a new release.

PR #987, opened 2024-10-26
Update configs and constants for Xcode 16

amorde:


ynyyn:
Note this update may potentially break CocoaPods downstream custom shell script **[CP] Embed Pods Frameworks** due to introducing `ENABLE_USER_SCRIPT_SANDBOXING = YES`. 

https://github.com/CocoaPods/CocoaPods/pull/12235

Issue #996, opened 2024-11-19
ArgumentError - [Xcodeproj] Unable to find compatibility version string for object version `70`.


mattgapadmi:
I am receiving this error after updating my .pbxproj file using Xcode 16.0:

> ArgumentError - [Xcodeproj] Unable to find compatibility version string for object version `70`.

I'm using xcodeproj 1.27.0

I noticed this in `lib/xcodeproj/constants.rb`:

```
COMPATIBILITY_VERSION_BY_OBJECT_VERSION = {
      77 => 'Xcode 16.0',
```

This doesn't seem right. I thought 70 is Xcode 16.0 and 77 is Xcode 16.1.

Any help would be appreciated.

DhanushKumar19:
I got this error on `pod install` when I created a directory as a folder (e.g., a Grey Blue Icon) within the project. Converting the folder to a group (e.g., a Grey color icon) resolved the issue for me.

faresbh:
Converting the folder to a group works thank you @DhanushKumar19

jerodji:
it works for me , thanks @DhanushKumar19

RomanPodymov:
If you cannot migrate from folders to groups, just change the project file version to Xcode 16

<img width="269" alt="Image" src="https://github.com/user-attachments/assets/122acf6d-d93c-40f0-96bf-04d12a92d053" />

videni:
@RomanPodymov ， where is the mysterious "Project Form ..." please?

dsrees:
@videni Select your project from the Project Navigator (⌘1) and then in the File Inspector (⌘⌥0), you'll see a Project Document section with a Project Format dropdown.

videni:
@dsrees  thanks, buddy.

flypigrmvb:
> 如果无法从文件夹迁移到组，只需将项目文件版本更改为 Xcode 16

古德

PR #1007, opened 2025-02-18
fix: Fixed xcode 16 compatibility

sprzenus:
Fixed issue #996


dbor-saxo:
I wonder if this will get merged :|

barrettj:
This PR has been open since February 2025 with no maintainer response. The fix is straightforward - just adding object version mappings to constants.rb.

  Can we get some clarity on whether this will be reviewed and merged? Many teams are blocked by this Xcode 16+ incompatibility and are relying on manual workarounds (downgrading objectVersion via sed).

  If Xcodeproj is no longer actively maintained, it would be helpful to know so the community can plan accordingly.

PR #1012, opened 2025-04-15
Remove empty `()` from PBXFileSystemSynchronizedRootGroup exceptions output

yimajo:
## Background

When using "Convert to Folder" on groups in projects within a CocoaPods workspace, an empty `exceptions = ();` entry is unnecessarily added to the pbxproj file.

## Why This Is a Problem

There's no need to include this empty configuration when there are no exceptions. If it were truly necessary, Xcode itself would add `exceptions = ();` when performing the Convert to Folder operation. These unnecessary entries make the project file needlessly complex and create unwanted diffs.

## Solution

The `PBXNativeTarget` class in this project already overrides the `to_hash_as` method to exclude empty collections (like packageProductDependencies) from serialization.

By extending the `PBXFileSystemSynchronizedRootGroup` class with a similar approach, we can prevent empty exceptions collections from being serialized, eliminating the unnecessary `exceptions = ();` entries in pbxproj files.

### DIFF
--- a/lib/xcodeproj/project/object/file_system_synchronized_root_group.rb
+++ b/lib/xcodeproj/project/object/file_system_synchronized_root_group.rb
@@ -68,6 +68,21 @@ module Xcodeproj
           return path if path
           super
         end
+
+        def to_hash_as(method = :to_hash)
+          hash_as = super
+          excluded_keys_for_serialization_when_empty.each do |key|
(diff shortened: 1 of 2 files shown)


adellibovi:
Thanks for opening this up, we are experiencing the same issue

PR #1037, opened 2026-05-06
Add compatibility version 70 for Xcode 26.0

nelsonng2002:
Fix ArgumentError - [Xcodeproj] Unable to find compatibility version string for object version `70` error


dnkoutso (approved):

iBotPeaches (commented):
`70` is already assigned to 16.0 though?

dnkoutso (changes requested):
yes this is not right

PR #1049, opened 2026-08-08
Support Swift Package Traits

PSchmiedmayer:
### Current situation & Problem

Xcodeproj treats Swift package traits as unknown attributes and removes them when saving a project.

### Release Notes

- Add trait support to remote and local Swift package references.
- Add serialization and deserialization coverage.

### Testing

- Verified that saving and reopening a project preserves its enabled traits


PSchmiedmayer:
@amorde Let us know if this is of interest to be merged; we have now encountered this in two projects as Traits become more used and we started to adopt them as well. Let me know if you have any feedback; thanks for maintaining the package!

PR #1050, opened 2026-09-01
Add support for Xcode 16.3+ shellScript format

### LINKED ISSUE #1017: Cannot handle projects from Xcode 16.3
The Xcode 16.3 project structure changes the shellScript of a build phase to an Array instead of string. Sample diff

```diff
- shellScript = "sh ../../scripts/swiftlint.sh\n";
+ shellScript = (
+  "sh ../../scripts/swiftlint.sh",
+  "",
+ ); 
```

This results in the error: **[Xcodeproj] Type checking error: got `Array` for attribute: attribute 'shellScript' (type: `simple`, classes: `[String]`, owner class : `PBXShellScriptBuildPhase`**

I haven't found any project-specific conditionals for declarations so I'm unsure what's the best (backwards compatible) way to solve this but the relevant line is [here](https://github.com/CocoaPods/Xcodeproj/blob/ab3dfa504b5a97cae3a653a8924f4616dcaa062e/lib/xcodeproj/project/object/build_phase.rb#L328).

Can someone point me in the right direction how I could fix this then I'd open a PR.

mt00chikin:
Xcode 16.3 and later stores multiline shell scripts as an array of strings, which breaks Xcodeproj's parsing, since PBXShellScriptBuildPhase's shell_script attribute is a String.

Running Xcodeproj on a project that has this format produces the following error:
`[Xcodeproj] Type checking error: got `Array` for attribute: Attribute `shellScript` (type: `simple`, classes: `[String]`, owner class: `PBXShellScriptBuildPhase`) (RuntimeError)`

This fixes the issue by aliasing the shell_script attribute. To ensure backwards compatibility, it is still represented as a String, joined by newline characters if the raw value in the project is an Array

This resolves https://github.com/CocoaPods/Xcodeproj/issues/1017


iBotPeaches:
I guess downside here if I understand it - once you move to arrays, Xcodeproj will just move you back to strings. I wonder if we can keep the format the source file is using. Since I imagine we might affect tooling if we save back a different format than user had it at.

mt00chikin:
> I guess downside here if I understand it - once you move to arrays, Xcodeproj will just move you back to strings. I wonder if we can keep the format the source file is using. Since I imagine we might affect tooling if we save back a different format than user had it at.

Agreed, this was also something I had considered as a potential alternative. As-is, the write will use the old string format, which Xcode will immediately attempt to convert back to an array representation that splits each newline into a separate element. It works, but it can be a bit noisy.
I wanted to keep the intermediate representation the same for backwards compatibility and ease, but I could update this so that the objectVersion defines the structure on write to match what is expected in Xcode. The parsing and writing functionality would use a different strategy based on the Xcode project format.

If you think going the route of Xcode objectVersion-specific mapping is preferable, I'm ok with making that update! I just wanted to start with minimal changes just to get this working. Would be great to update our project to Xcode 26 and continue to use fastlane tooling like increment_version_number which is built on top of this.

